PR #38, opened 2020-04-30
Graql Delete Scenarios

flyingsilverfin:
## Goals
Create comprehensive tests for delete functionality

## Changes
* Define and flesh out scenarios for deletion
* Rename background steps in other `language/` feature files

### DIFF
--- a/behaviour/graql/language/delete.feature
+++ b/behaviour/graql/language/delete.feature
@@ -15,13 +15,17 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 Feature: Graql Delete Query
-
-  Scenario: deleted instance no longer visible as instance of its type
-
-  Scenario: deleted role player is not part of relation
-
-  Scenario: deleted attribute not visible from any owner of attribute
-
-  Scenario: deleted attribute owning instance not visible from attributes
-
-  Scenario: deleting attribute ownership retains attribute and owner but no mutual visibility
+  Scenario: delete an instance using 'thing' meta label succeeds
+  Scenario: delete an relation instance using 'relation' meta label succeeds
+  Scenario: delete an entity instance using 'entity' meta label succeeds
+  Scenario: delete an attribute instance using 'attribute' meta label succeeds
+  Scenario: delete a role player from a relation removes the player from the relation
+  Scenario: delete an instance the instance not a player in any relation anymore
+  Scenario: delete duplicate role players from a relation removes duplicate player from relation
+  Scenario: delete attribute ownership makes attribute invisible to owner
+  Scenario: delete role players in multiple statements are all deleted
+  Scenario: delete all role players of relation cleans up relation instance
+  Scenario: delete more role players than exist throws
+  Scenario: delete a role player with too-specific (downcasting) role label throws
+  Scenario: delete an instance using wrong type throws
+  Scenario: delete an instance using too-specific (downcasting) type throws
--- a/behaviour/graql/language/delete.feature
+++ b/behaviour/graql/language/delete.feature
@@ -266,7 +266,7 @@ Feature: Graql Delete Query
       match (friend: $x, friend: $y) isa friendship; get;
       """
     Then uniquely identify answer concepts
-      | x    | y    |
+      | x    | y     |
       | BOB  | CAR   |
       | CAR  | BOB   |
 
@@ -771,24 +771,6 @@ Feature: Graql Delete Query
         $refl (friend: $x, friend: $x) isa friendship, has ref $r1;
         $f1 (friend: $x, friend: $y) isa friendship, has ref $r2;
       """
-
-    Then get answers of graql query
-      """
-      match $f (friend: $x) isa friendship; get;
-      """
-    Then uniquely identify answer concepts
-      | f     | x     |
-      | F2    | ALEX  |
-      | F2    | JOHN  |
-
-    Then get answers of graql query
-      """
-      match $n isa name; get;
-      """
-    Then uniquely identify answer concepts
-      | n     |
-      | nJHN  |
-      | nALX  |
 
     Then get answers of graql query
       """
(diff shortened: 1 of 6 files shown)


behaviour/graql/language/delete.feature
```diff
@@ -16,12 +33,566 @@
...
+    Then get answers of graql query
+      """
+      match $x isa person; get;
+      """
+    Then uniquely identify answer concepts
+      | x  |
+      | P2 |
+
+
+  Scenario: delete an relation instance using 'relation' meta label succeeds
```
jmsfltchr, line 143: `a relation`

behaviour/graql/language/delete.feature
```diff
@@ -16,12 +33,566 @@
...
+
+    Then get answers of graql query
+      """
+      match $r (friend: $x) isa friendship; get;
+      """
+    Then uniquely identify answer concepts
+      | r     | x    |
+      | FR1   | P2   |
+      | FR2   | P3   |
+
```
jmsfltchr, line 378: Should this also check that the `person` was deleted (not just removed as a roleplayer)?

behaviour/graql/language/delete.feature
```diff
@@ -16,12 +33,566 @@
...
+      """
+      insert
+      $x isa person, has ref 0;
+      $y isa person, has ref 1;
+      $r (friend: $x, friend: $x, friend: $y) isa friendship, has ref 2;
+      """
+    When the integrity is validated
+
+    Then concept identifiers are
+      |      | check | value     |
```
jmsfltchr, line 391: Overall I'm finding using `ref 0, 1, 2` a bit hard to read - I suggest we use something symbolic - `ALI`, `BOB`, `CHR` for instance across the tests. Doesn't have to be done now. At the moment though the misaligned `P1` -> `ref:0` has thrown me off about 3 times haha

behaviour/graql/language/delete.feature
```diff
@@ -16,12 +33,566 @@
...
+      | P2   | key   | ref:1     |
+      | FR   | key   | ref:2     |
+
+    Then graql delete
+      """
+      match
+        $r (friend: $x, friend: $x) isa friendship;
+      delete
+        $r (friend: $x, friend: $x);
+      """
```
jmsfltchr, line 402: Do we have the case for 
```
match
$r (friend: $x, friend: $x) isa friendship;
delete
$r (friend: $x);
```
?

jmsfltchr: Also this one:
```
match
$r (friend: $x) isa friendship;
delete
$r (friend: $x);
```
But where `$x` plays friend twice as in this example

behaviour/graql/language/delete.feature
```diff
@@ -16,12 +33,566 @@
...
+      match
+        $r (friend: $x, friend: $y, friend: $z) isa friendship;
+        $x isa person, has ref 0;
+        $y isa person, has ref 1;
+        $z isa person, has ref 2;
+      delete
+        $r (friend: $x);
+        $r (friend: $y);
+      """
+
```
jmsfltchr, line 442: Nice

behaviour/graql/language/delete.feature
```diff
@@ -16,12 +33,566 @@
...
+    Then answer size is: 0
 
-  Scenario: deleted attribute owning instance not visible from attributes
 
-  Scenario: deleting attribute ownership retains attribute and owner but no mutual visibility
+  Scenario: using unmatched variable in delete throws even without data
+    Then graql delete throws
+      """
+      match $x isa person; delete $n isa name;
+      """
```
jmsfltchr, line 598: In the case that we can delete attribute ownership we should have a scenario (or more) for deleting shared attributes, ensuring that the other owners are unaffected. 
If we are to release then we need to test that this can be achieved presently, presumably by deleting implicit relations. Depends if they are gone by next release. Either way, we could have the tests and `@ignore` them. 
Also we should have a scenario for deleting the last ownership of an attribute - is the attribute itself queryable or not after that?

behaviour/graql/language/delete.feature
```diff
@@ -16,12 +33,566 @@
...
+    Then graql delete throws
+      """
+      match
+        $x isa person, has ref 0;
+        $y isa person, has ref 1;
+        $r (friend: $x, friend: $y) isa friendship;
+      delete
+        $r (special-friend: $x);
+      """
+
```
jmsfltchr, line 545: I think we should also have some cases for deleting larger patterns in case something unexpected happens

behaviour/graql/language/delete.feature
flyingsilverfin: sorted, using names for people and starting number at 1,2... for relations

behaviour/graql/language/delete.feature
flyingsilverfin: can do, why not

behaviour/graql/language/delete.feature
flyingsilverfin: good ideas!

flyingsilverfin:
TODO:
* scenario when removing ownership of an attribuet owned by multiple owners
* scenarios with keys

behaviour/graql/language/delete.feature
flyingsilverfin: this is actually causing an issue :P

behaviour/graql/language/delete.feature
flyingsilverfin: aight

behaviour/graql/language/delete.feature
```diff
@@ -16,12 +16,728 @@
...
+      """
+      match
+        $x isa person;
+        $r isa name; $r "John";
+      delete
+        $r isa person;
+      """
+
+
+  Scenario: delete an instance using too-specific (downcasting) type throws
```
jmsfltchr, line 209: We should also test that deleting types throws!

flyingsilverfin:
@jmsfltchr  ok, should be good finally!

PR #43, opened 2020-05-19
Remove all usages of implicit concepts

flyingsilverfin:
According to https://github.com/graknlabs/grakn/issues/5614, we aim to remove implicit relations from Grakn. This means usages in `verification` referring to implicit relations or roles need to be modified or removed.


haikalpribadi:
I'll review after @jmsfltchr . Just nudge me when it's ready

flyingsilverfin:
@haikalpribadi  james has finished his review :)

Issue #18, opened 2020-03-03
Complete Axiomatic Tests - implement missing Graql query scenarios


flyingsilverfin:
The BDD scenarios currently cover relatively comprehensively (except verification/errors that should throw as in #11):
* Undefine
* Define
* insert
* match
* delete

Some missing tests that should be implemented have the scenario names written in the respective feature files: match, get, insert all have missing implementations.

I will supervise as needed when completing this task

alexjpwalker:
The implementation of missing scenarios is complete

PR #48, opened 2020-06-02
New graql language tests: define and mutate

alexjpwalker:
The goal of this PR is to add new define and mutate tests for the graql language.

This PR forms part of the work towards #18.


behaviour/graql/language/mutate.feature
```diff
@@ -0,0 +1,695 @@
...
+  @ignore
+  # TODO: re-enable when rules cannot infer abstract relations
+  Scenario: change concrete relation type to abstract throws on commit if it is the conclusion of any rule
+
+  @ignore
+  # TODO: check if rules can infer abstract attributes
+  Scenario: change concrete attribute type to abstract throws on commit if it is the conclusion of any rule
+
+  ###############
+  # INHERITANCE #
```
flyingsilverfin, line 396: In my mind, "inheritance" has to do with the fact that a subtype gets whatever the supertype has - do all these tests mutate hierarchies and then check that properties are preserved downards?

If not, maybe a better description is "Hierarchy Mutation"?

behaviour/graql/language/mutate.feature
```diff
@@ -0,0 +1,695 @@
...
+  Scenario: assign new super-attribute throws if it has a different value type to the current one (?)
+
+  # TODO: write this if 'assign new super-attribute throws if it has a different value type ..' turns out to not throw
+  Scenario: assign new super-attribute throws if it has existing data and a different value type to the new supertype (?)
+
+  # TODO: write this once 'define new `sub` on attribute type changes its supertype' passes
+  Scenario: assign new super-attribute throws if new supertype has a regex and existing data doesn't match it (?)
+
+
+  Scenario: define additional 'plays' is visible from all children
```
flyingsilverfin, line 585: these are a better fit for "Inheritance", what do you think?

behaviour/graql/language/undefine.feature
```diff
@@ -20,7 +20,7 @@ Feature: Graql Undefine Query
     Given connection has been opened
     Given connection delete all keyspaces
     Given connection open sessions for keyspaces:
-      | test_define |
+      | test_undefine |
```
flyingsilverfin, line 23: good catch!!

behaviour/graql/language/define.feature
```diff
@@ -230,157 +271,79 @@ Feature: Graql Define Query
...
-    Then concept identifiers are
-      |     | check | value  |
-      | ANI | label | animal |
-    Then uniquely identify answer concepts
-      | x   |
-      | ANI |
 
 
-  Scenario: define concrete subtype of abstract entity creates child of its parent type
+  Scenario: define multiple copies of the same 'plays' creates a type that plays that role
```
flyingsilverfin, line 281: we like the term "idempotent"

behaviour/graql/language/define.feature
```diff
@@ -230,157 +271,79 @@ Feature: Graql Define Query
...
+      | HOU | label | house |
     Then uniquely identify answer concepts
       | x   |
-      | ANI |
-      | HOR |
+      | HOU |
 
 
-  Scenario: define abstract subtype of abstract entity creates child of its parent type
+  Scenario: define multiple copies of the same 'has' creates a type that has that attribute
```
flyingsilverfin, line 302: here again

behaviour/graql/language/define.feature
```diff
@@ -230,157 +271,79 @@ Feature: Graql Define Query
...
 
 
-  Scenario: define additional 'plays' is visible from all children
-    Given graql define
-      """
-      define employment sub relation, relates employer;
-      """
-    Given the integrity is validated
-
+  Scenario: define multiple copies of the same 'key' creates a type that has that key
```
flyingsilverfin, line 322: here too

behaviour/graql/language/define.feature
```diff
@@ -636,112 +598,61 @@ Feature: Graql Define Query
       | FAA |
 
 
-  Scenario: define abstract relation type creates a type
+  Scenario: define abstract relation type with no roleplayers creates a type
```
flyingsilverfin, line 601: maybe call this one "is valid" at the end, instead of "creates a type", is more descriptive

behaviour/graql/language/define.feature
```diff
@@ -636,112 +598,61 @@ Feature: Graql Define Query
...
+      | CON | label | connection |
     Then uniquely identify answer concepts
       | x   |
-      | MEM |
-      | GYM |
+      | CON |
 
 
-  Scenario: define abstract subtype of abstract relation creates child of its parent type
+  Scenario: define relation type with multiple copies of the same 'relates' creates a relation type with all distinct roleplayers
```
flyingsilverfin, line 619: eh? this title no make sense to me

behaviour/graql/language/define.feature
```diff
@@ -636,112 +598,61 @@ Feature: Graql Define Query
...
-  @ignore
-  # re-enable when we can inherit 'relates
-  Scenario: define additional 'relates' is visible from all children
+  Scenario: define relation type 'relates' to a role it plays itself creates a relation type
     Given graql define
       """
       define
-      part-time-employment sub employment;
-      employment sub relation, relates employer;
+      recursive-function sub relation, relates function, plays function;
```
flyingsilverfin, line 643: haha!! yess!!

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,700 @@ Feature: Graql Define Query
       | VDR |
 
 
-  Scenario: define abstract attribute type creates a type
+  Scenario: define attribute type that has itself creates an attribute
```
flyingsilverfin, line 931: better description might be "defining an attribute owning itself creates self-ownership" - it's not about the attribute, it's about the ownership that is interesting

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,700 @@ Feature: Graql Define Query
...
     Then uniquely identify answer concepts
       | x   |
-      | NOL |
-      | NLE |
+      | BOB |
+      | RUL |
 
 
-  Scenario: define abstract subtype of abstract attribute creates child of its parent type
+  Scenario: define a rule that infers a relation creates a rule
```
flyingsilverfin, line 981: i kind of like having the word "valid" in these titles - "creates a valid rule" lets the reader immediately see that the point is here to have a "valid" result :)

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,700 @@ Feature: Graql Define Query
...
+      |     | check | value              |
+      | HAI | label | haikal-is-employed |
+      | RUL | label | rule               |
+    Then uniquely identify answer concepts
+      | x   |
+      | HAI |
+      | RUL |
 
+
+  Scenario: define rule that infers a `key` creates a rule
```
flyingsilverfin, line 1007: basically "creates a rule" on its own seems like a vacuous statement kinda?

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,700 @@ Feature: Graql Define Query
...
+      | ONL | label | only-child-rule |
+      | RUL | label | rule            |
+    Then uniquely identify answer concepts
+      | x   |
+      | ONL |
+      | RUL |
+
+
+  # TODO: better scenario description? what exactly does 'unbound negated pattern variable' mean?
+  Scenario: define a rule with unbound negated pattern variables throws
```
flyingsilverfin, line 1120: your TODO comment - all cleared up? "unbound negated pattern variable" means that the negation contains at least one variable that is bound outside of the negation sub-clause as well

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,700 @@ Feature: Graql Define Query
...
+      }, then {
+        $register has has-robert false;
+      };
+      """
+
 
-  Scenario: define a rule causing a loop throws on commit (eg. conclusion is negated in the `when`)
+  @ignore
+  # TODO: re-enable when all rules with nested negations throw on commit
+  Scenario: define a rule with nested negation throws on commit
```
flyingsilverfin, line 1140: this isn't expected to throw - not(not()) is fine (we normalise all the queries to disjunctive normal form, so we flatten the double negated form), not($x has...; not ($x... $a)) should throw though

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,700 @@ Feature: Graql Define Query
...
+      i-did-a-bad-typo sub rule,
+      when {
+        $p has name "I am a person";
+      }, then {
+        $q has nickname "Who am I?";
+      };
+      """
+
+
+  Scenario: define rule with an undefined attribute set in `then` throws on commit
```
flyingsilverfin, line 1254: waaaait this throws??

flyingsilverfin:
I left some comments, some general, some applicable to all of them

behaviour/graql/language/mutate.feature
```diff
@@ -0,0 +1,695 @@
...
+
+      name sub attribute, value string;
+      email sub attribute, value string;
+      start-date sub attribute, value datetime;
+      """
+    Given the integrity is validated
+
+
+  ##############
+  # PRIMITIVES #
```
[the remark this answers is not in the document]
alexjpwalker: Yeah to be honest I couldn't think of a good name for this section. The best alternative was "General".
Actually I think "Concepts" is what I meant to say. Basic entity/relation/attr, rules, and has/plays/relates tests.

behaviour/graql/language/mutate.feature
```diff
@@ -0,0 +1,695 @@
...
+
+
+  Scenario: change attribute type to entity type throws
+    Then graql define throws
+      """
+      define name sub entity;
+      """
+
+
+  Scenario: define additional 'has' on a type adds attribute to it
```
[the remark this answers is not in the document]
alexjpwalker: All the tests passed in my latest local run, including this one.

behaviour/graql/language/mutate.feature
```diff
@@ -0,0 +1,695 @@
...
+      $y isa product, has name "Ham";
+      """
+    Given the integrity is validated
+    When graql define
+      """
+      define
+      product key barcode;
+      """
+    When the integrity is validated
+    When get answers of graql query
```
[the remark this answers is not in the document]
alexjpwalker: Ah yes I should've enquired about this one. Will follow up in Discord.

behaviour/graql/language/mutate.feature
alexjpwalker: Originally it was called "Schema Mutation: Inheritance", then I moved the tests to this file and dropped the prefix. I agree that Hierarchy Mutation is a better name.

behaviour/graql/language/mutate.feature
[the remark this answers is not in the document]
alexjpwalker: Updated to Concepts.

behaviour/graql/language/mutate.feature
```diff
@@ -0,0 +1,695 @@
...
+      start-date sub attribute, value datetime;
+      """
+    Given the integrity is validated
+
+
+  ##############
+  # PRIMITIVES #
+  ##############
+
+  Scenario: repeatedly defining existing type keeps its properties intact
```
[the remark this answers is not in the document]
alexjpwalker: Renamed to: `re-define existing type keeps its properties intact and is idempotent`

behaviour/graql/language/mutate.feature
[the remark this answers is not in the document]
alexjpwalker: As discussed on Discord, we should talk about this tomorrow [03 Jun]

behaviour/graql/language/mutate.feature
alexjpwalker: Renamed to Hierarchy Mutation.

behaviour/graql/language/mutate.feature
alexjpwalker: Added "Inheritance" section only containing these four tests.

behaviour/graql/language/define.feature
alexjpwalker: Renamed to `define 'plays' is idempotent`

behaviour/graql/language/define.feature
alexjpwalker: Renamed to `define 'has' is idempotent`

behaviour/graql/language/define.feature
alexjpwalker: Renamed to `define 'key' is idempotent`

behaviour/graql/language/define.feature
alexjpwalker: To be honest, quite a lot of the tests could be renamed to include 'is valid', I just find the words a bit redundant. I prefer to maintain some consistency by specifying:
1- the action "define abstract relation type with no roleplayers"
2- the outcome "creates a type"

But in this case I suppose 'creates a type' is no better, giving across the meaning of 'creating a type creates a type'. It's a tough one really. But I definitely have an aversion to writing "is valid" as the outcome - to me that's not a valid outcome.

I think I'll go with `define relation type with no roleplayers is possible when marked as abstract`, emphasising that the outcome is you've defined a relation type with no roleplayers.

alexjpwalker: Or even `is valid when marked as abstract`, heh

behaviour/graql/language/define.feature
alexjpwalker: I basically meant: _when you define a relation type with duplicate 'relates', you get a relation type that has all of the specified 'relates' as roleplayers, without any duplicates_.
It's hard to put it into a concise description.

behaviour/graql/language/define.feature
alexjpwalker: Yes I agree, and to be honest there are quite a number of cases where I've written "creates a relation" or "creates an attribute" as the outcome because it's what actually happens as a result of the _define_, but you're right that it's not the interesting part. I will go through and address them.

behaviour/graql/language/define.feature
alexjpwalker: Renamed to `when defining a relation type, duplicate 'relates' are idempotent`.

behaviour/graql/language/define.feature
alexjpwalker: Yeah I'm no big fan of "creating a rule creates a rule" either.
I think I might change these to the format: `a rule can infer a relation`.
I was kinda obsessed with including the word "define" in every test, but it's in the filename for god's sakes.

behaviour/graql/language/define.feature
alexjpwalker: Changed to `a rule can infer a key`.

behaviour/graql/language/define.feature
alexjpwalker: Changed to `a rule can infer a relation`.

behaviour/graql/language/define.feature
alexjpwalker: Hmm, I don't see any variable in this negation that is bound outside of the negation block - do you mean that the negation **must** contain at least one variable that was bound outside of the negation block?

behaviour/graql/language/define.feature
alexjpwalker: Okay, here's a new case that should throw.
```
define
      nickname sub name;
      person has nickname;
      unemployed-robert-maybe-doesnt-not-have-nickname-bob sub rule,
      when {
        $p isa person;
        not {
          (employee: $p) isa employment;
          not {
            $p has name "Robert";
          };
        };
      }, then {
        $p has nickname "Bob";
      };```

File: behaviour/graql/language/define.feature (hunk 15 above)
Comment (alexjpwalker) [reply behaviour/graql/language/define.feature:1254]: Yes. It should, shouldn't it? `person` doesn't have an attribute called `age` so it's an invalid rule

**Comment by alexjpwalker** [comment]
Thank you for the feedback. Fixed issues raised in PR review. Let me know if anything still looks bad.

File: behaviour/graql/language/mutate.feature (hunk 16 above)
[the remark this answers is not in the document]
Comment (flyingsilverfin) [reply behaviour/graql/language/mutate.feature]: cool

File: behaviour/graql/language/mutate.feature (hunk 19 above)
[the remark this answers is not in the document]
Comment (flyingsilverfin) [reply behaviour/graql/language/mutate.feature]: yep!

File: behaviour/graql/language/mutate.feature (hunk 1 above)
Comment (flyingsilverfin) [reply behaviour/graql/language/mutate.feature]: cool

File: behaviour/graql/language/mutate.feature (hunk 2 above)
Comment (flyingsilverfin) [reply behaviour/graql/language/mutate.feature:588]: great :)

File: behaviour/graql/language/define.feature (hunk 5 above)
Comment (flyingsilverfin) [reply behaviour/graql/language/define.feature]: yas

File: behaviour/graql/language/define.feature (hunk 8 above)
Comment (flyingsilverfin) [reply behaviour/graql/language/define.feature]: yas this makes sense :)

File: behaviour/graql/language/define.feature (hunk 13 above)
Comment (flyingsilverfin) [reply behaviour/graql/language/define.feature:1120]: yes - MUST

File: behaviour/graql/language/define.feature (hunk 15 above)
Comment (flyingsilverfin) [reply behaviour/graql/language/define.feature:1249]: ah ok, cool

File: behaviour/graql/language/mutate.feature (hunk 17 above)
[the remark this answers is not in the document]
Comment (flyingsilverfin) [reply behaviour/graql/language/mutate.feature:87]: i thought for a minute that `match $x has name; get;` on the schema level doesn't return anything, but the real difference is `match $x has $y; get;` on the schema level (where $x and $y are types) -- _this_ query doesn't run!

File: behaviour/graql/language/define.feature (hunk 7 above)
Comment (flyingsilverfin) [reply behaviour/graql/language/define.feature]: lets discuss this on a call

File: behaviour/graql/language/mutate.feature (hunk 18 above)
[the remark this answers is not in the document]
Comment (flyingsilverfin) [reply behaviour/graql/language/mutate.feature:168]: lets discuss this on a call

File: behaviour/graql/language/define.feature (hunk 14 above)
Comment (alexjpwalker) [reply behaviour/graql/language/define.feature:1138]: Just documenting the results of the call

- Unwanted error most likely triggered because `nickname sub name`, could be fixed by making `nickname sub attribute, value string`

File: behaviour/graql/language/define.feature (hunk 13 above)
Comment (alexjpwalker) [reply behaviour/graql/language/define.feature:1120]: Renamed to `define a rule with a negation block whose pattern variables are all unbound outside the negation block throws`

File: behaviour/graql/language/define.feature (hunk 14 above)
Comment (alexjpwalker) [reply behaviour/graql/language/define.feature:1138]: Updated to `nickname sub attribute, value string`.

The new error is:
```
INVALID_ARGUMENT: InvalidKBException-A structural validation error has occurred. Please correct the [`1`] errors found. 
The rule [unemployed-robert-maybe-doesnt-not-have-nickname-bob] contains a nested negation block
```
which is the correct behaviour!

File: behaviour/graql/language/mutate.feature (hunk 18 above)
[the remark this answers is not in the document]
Comment (alexjpwalker) [reply behaviour/graql/language/mutate.feature:168]: Just documenting the results of the Discord chat:
- This scenario will be split up into two scenarios, the current one (which should throw and will be @ignored for now), and a new one which `insert`s the keys prior to commit, and should pass with a successful commit.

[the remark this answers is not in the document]
Comment (alexjpwalker) [reply behaviour/graql/language/mutate.feature]: And it's done - `additional 'key' can be defined on a type, as long as existing instances have correct keys prior to commit` is now implemented!

File: behaviour/graql/language/define.feature [hunk 20]
```diff
@@ -1018,92 +928,695 @@ Feature: Graql Define Query
...
+  Scenario: define abstract subtype of concrete entity throws an error
+    Then graql define throws
+      """
+      define
+      exception sub entity;
+      grakn-exception sub exception, abstract;
+      """
+
+
+  Scenario: repeatedly define type as abstract creates an abstract type
```
jmsfltchr, line 1606: Should this be called `defining type as abstract is idempotent`?

behaviour/graql/language/mutate.feature
```diff
@@ -0,0 +1,698 @@
...
+      $x isa person, has name "Maria", has email "maria@grakn.ai";
+      """
+    Given the integrity is validated
+    Then graql define throws
+      """
+      define name regex "^A.*$";
+      """
+
+
+  Scenario: define additional 'regex' on a long-valued attribute type throws
```
jmsfltchr, line 250: The same goes for everything that isn't a string I suppose

behaviour/graql/language/define.feature
```diff
@@ -230,157 +271,79 @@ Feature: Graql Define Query
...
-      |       | check | value |
-      | CHILD | label | child |
-      | EMAIL | label | email |
-    Then uniquely identify answer concepts
-      | x     | y     |
-      | CHILD | EMAIL |
-      | CHILD | EMAIL |
 
 
   #############
```
jmsfltchr, line 349: I feel we're missing a section on attributes, including the simple stuff like being able to define an attribute of each type, and that each type of attribute can be owned. Then perhaps there are some more fundamental attribute cases I can't think of yet. This may be overkill, I'll leave it to you, but attributes often present difficulties, and with the multiple datatypes possible they, in some ways, present more opportunities for failure than entities and relations.

behaviour/graql/language/define.feature
```diff
@@ -82,6 +81,48 @@ Feature: Graql Define Query
       | CHD |
 
 
+  Scenario: define that a type 'has' something undefined throws
+    Then graql define throws
+      """
+      define book sub entity, has pages;
+      """
+
```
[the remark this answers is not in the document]
flyingsilverfin: @jmsfltchr  the `integrity is validated` step is currently a noop, i haven't finished it yet! It's currently planned for after the 2.0 integration!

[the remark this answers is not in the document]
jmsfltchr: I realise this, though I still feel we should add it in the tests where we want it to act so that we don't have to revisit them later :)

behaviour/graql/language/define.feature
alexjpwalker: Renamed to `a relation type can be defined with no roleplayers when it is marked as abstract`

behaviour/graql/language/define.feature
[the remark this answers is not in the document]
alexjpwalker: What is `the integrity is validated` actually going to do?
I've been stuffing it pretty much everywhere after every successful write transaction. Didn't think of putting it after those that are expected to fail.
Is it valuable? After all, these `graql define throws` are all transactions, and if they throw, they will certainly not have committed anything.

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,695 @@ Feature: Graql Define Query
...
+      |     | check | value              |
+      | HAI | label | haikal-is-employed |
+      | RUL | label | rule               |
+    Then uniquely identify answer concepts
+      | x   |
+      | HAI |
+      | RUL |
 
+
+  Scenario: a rule can infer a `key`
```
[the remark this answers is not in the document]
alexjpwalker: Replied in Discord # graql

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,695 @@ Feature: Graql Define Query
...
+        not {
+          $p isa person, has name "Robert";
+        };
+      }, then {
+        $register has has-robert false;
+      };
+      """
+
+
+  Scenario: define a rule with nested negation throws on commit
```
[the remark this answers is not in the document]
alexjpwalker: Done, and now I feel sorry for Robert, even though he's just a test case:
```
define
      nickname sub attribute, value string;
      residence sub relation, relates resident;
      person has nickname, plays resident;
      unemployed-homeless-robert-have-nickname-bob sub rule,
      when {
        $p isa person, has name "Robert";
        not {
          (employee: $p) isa employment;
        };
        not {
          (resident: $p) isa residence;
        };
      }, then {
        $p has nickname "Bob";
      };
```

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,695 @@ Feature: Graql Define Query
...
+      nickname sub name;
+      person has nickname;
+      """
+    Given the integrity is validated
+    Then graql define throws
+      """
+      define
+      sophie-and-fiona-have-nickname-fi sub rule,
+      when {
+        {$p has name "Sophie";} or {$p has name "Fiona";};
```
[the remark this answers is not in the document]
alexjpwalker: So the issue of the variable not being bound in the outer scope actually does not feature in the error message - however I can definitely see that it conceivably could/should.

Updated to include `$p isa person;`

behaviour/graql/language/define.feature
```diff
@@ -1018,92 +928,695 @@ Feature: Graql Define Query
...
+      bonnie-and-clyde-are-partners-in-crime sub rule,
+      when {
+        $bonnie isa person, has name "Bonnie";
+        $clyde isa person, has name "Clyde";
+      }, then {
+        (criminal: $bonnie, sidekick: $clyde) isa partners-in-crime;
+      };
+      """
+
+  Scenario: define rule that infers an abstract attribute throws on commit (?)
```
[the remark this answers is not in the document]
alexjpwalker: Right, so this one (along with a number of mutate tests) is unimplemented because I wanted to get on to migrating the grakn integration tests. And the (?) is because I don't currently know what the behaviour will be. My guess is it will, in fact, not throw on commit and turn out to be a bug.

But to be honest, it's not a lot of effort to write, might as well do it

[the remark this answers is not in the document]
alexjpwalker: Done
```
@ignore
  # TODO: re-enable when rules cannot infer abstract attribute values
  Scenario: define rule that infers an abstract attribute value throws on commit
    Then graql define throws
      """
      define
      number-of-devices sub attribute, value long, abstract;
      person has number-of-devices;
      karl-is-allergic-to-technology sub rule,
      when {
        $karl isa person, has name "Karl";
      }, then {
        $karl has number-of-devices 0;
      };
      """
```

behaviour/graql/language/define.feature
alexjpwalker: Renamed.

behaviour/graql/language/mutate.feature
alexjpwalker: Yes. But I question whether it's really worth having a test for each individual attribute type. I suppose one could argue that longs and doubles might support regexes in the future?

Nonetheless, not sure how important this is

behaviour/graql/language/define.feature
alexjpwalker: I think you're right that some of the basic stuff should be tested individually against each attribute type.

I had the same decisions to make when it came to test duplication across entity/relation/attribute. "Is this test significant enough to justify rewriting it for each individual kind of type?" I duplicated the really fundamental cases, like "a type can be marked as `abstract`", and "a type inherits `plays` from its parent"; and of course, the all-important "defining a (entity|relation|attribute) creates a type". So it probably makes sense to run through the very basics of each attribute value type as well.

alexjpwalker: Added:

```
an attribute with type `string` can be defined
an attribute with type `long` can be defined
an attribute with type `double` can be defined
an attribute with type `boolean` can be defined
an attribute with type `datetime` can be defined
a type can `has` an attribute of type `string`
a type can `has` an attribute of type `long`
a type can `has` an attribute of type `double`
a type can `has` an attribute of type `boolean`
a type can `has` an attribute of type `datetime`
```

behaviour/graql/language/define.feature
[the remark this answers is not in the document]
jmsfltchr: Poor Robert indeed, though thankfully for him, this is flagged as an error

behaviour/graql/language/mutate.feature
jmsfltchr: I leave it to you to judge!

behaviour/graql/language/define.feature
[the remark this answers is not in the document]
alexjpwalker: We have decided to add it after all `graql define throws`.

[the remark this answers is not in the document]
alexjpwalker: Done!

PR #52, opened 2020-06-09
Add Graql Get, Match and Undefine scenario drafts

alexjpwalker:

To have a draft specification of Graql Get, Match and Undefine.

This PR forms part of the work towards #18

### DIFF
--- a/behaviour/graql/language/match.feature
+++ b/behaviour/graql/language/match.feature
@@ -28,8 +28,7 @@ Feature: Graql Match Clause
 
   Scenario: `sub` matches the specified type and all its subtypes
 
-  # TODO: ????
-  Scenario: `sub!` matches ??? (?)
+  Scenario: `sub!` matches the specified type and its direct subtypes
 
   # TODO: ensure there are types that don't have the specified attribute, as well, so we know it's a precise match
   Scenario: `has` matches types that have the specified attribute type
--- a/behaviour/graql/language/match.feature
+++ b/behaviour/graql/language/match.feature
@@ -24,7 +24,65 @@ Feature: Graql Match Clause
     Given transaction is initialised
     Given the integrity is validated
 
-  Scenario: Disjunctions return the union of composing query statements
+  Scenario: `type` matches only the specified type, and does not match subtypes
+
+  Scenario: `sub` matches the specified type and all its subtypes
+
+  # TODO: ????
+  Scenario: `sub!` matches ??? (?)
+
+  # TODO: ensure there are types that don't have the specified attribute, as well, so we know it's a precise match
+  Scenario: `has` matches types that have the specified attribute type
+
+  Scenario: `has` matches types that have only a subtype of the specified attribute type (?)
+
+  Scenario: `plays` matches types that can play the specified role
+
+  Scenario: `key` matches types that have the specified attribute type as a key (?)
+
+  Scenario: `has` matches types that have the specified attribute type, even if they use it as a key (?)
+
+  Scenario: `relates` matches types that have the specified roleplayer, regardless of their role
+
+  Scenario: `relates` with `as` matches only types that have the specified roleplayer playing the specified role
+
+  Scenario: `relates` matches/does not match types that block the specified roleplayer with `as` (?)
+
+  Scenario: `isa` matches things of the specified type and all its subtypes
+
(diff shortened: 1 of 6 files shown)


behaviour/graql/language/match.feature
```diff
@@ -24,7 +24,65 @@ Feature: Graql Match Clause
     Given transaction is initialised
     Given the integrity is validated
 
-  Scenario: Disjunctions return the union of composing query statements
+  Scenario: `type` matches only the specified type, and does not match subtypes
+
+  Scenario: `sub` matches the specified type and all its subtypes
+
+  # TODO: ????
+  Scenario: `sub!` matches ??? (?)
```
flyingsilverfin, line 32: what's this?

behaviour/graql/language/match.feature
```diff
@@ -24,7 +24,65 @@ Feature: Graql Match Clause
...
+
+  Scenario: disjunctions return the union of composing query statements
+
+  Scenario: a negation matches if the negated block has no matches
+
+  Scenario: a negation does not match if the negated block has any matches
+
+  Scenario: `has` matches attribute ownerships inferred from rules
+
+  Scenario: `isa` matches relations inferred from rules
```
flyingsilverfin, line 85: I'm wondering if we want to have a whole separate file for testing reasoner capabilities, because there are some more:
1. appending role players
2. adding a "second" supertype of an instance (this is a sneaky reasoning thing we can do, not documented anywhere ;), maybe we can get kasper who wrote the entire reasoning engine to review the set of reasoning scenarios in the future)
3. creating new attribute and an ownership thereof in the `then` of the rule..


there are more, i just think that is a good  separate scope to have a new file for.

what do you think?

flyingsilverfin: there's even a `reasoning` package in verification/behaviour already i think~

behaviour/graql/language/undefine.feature
```diff
@@ -270,3 +281,24 @@ Feature: Graql Undefine Query
...
+
+  Scenario: a type can be undefined after deleting all of its instances
+
+  Scenario: undefine 'plays' from type throws if there is an instance playing that role in an existing relation
+
+  Scenario: 'relates' can be undefined when there are existing relations, but none of them have that roleplayer
+
+  Scenario: undefine 'relates' from relation type throws if there is an existing instance with that roleplayer
+
+  Scenario: 'has' can be undefined when there are existing instances, but none of them have that attribute
```
flyingsilverfin, line 300: i generally like to use the term "ownership" or "owns" for having an attribute, for future reference, if you think that is also a good term to use!

behaviour/graql/language/match.feature
alexjpwalker: Updated to: `sub! matches the specified type and its direct subtypes`

behaviour/graql/language/undefine.feature
alexjpwalker: Yeah I'm also seeing it used extensively in the ITs since starting to migrate them

behaviour/graql/language/match.feature
alexjpwalker: Ok I will run the name `reasoner/inference.feature` past Haikal.

alexjpwalker: We're thinking `resolution.feature` (which might change in the future as more tests are added)

behaviour/graql/language/undefine.feature
alexjpwalker: Updated to
```
Scenario: attribute ownership can be undefined when there are instances of the owner, but none of them own that attribute

Scenario: undefine attribute ownership throws if any instance of the owner currently owns that attribute```

File: behaviour/graql/reasoner/resolution.feature [hunk 4]
```diff
@@ -0,0 +1,20 @@
...
+    Given transaction is initialised
+
+  Scenario: `has` matches inferred attribute ownerships
+
+  Scenario: `isa` matches inferred relations
+
+  Scenario: `isa` matches inferred roleplayers in relation instances
+
+  # TODO: should this be allowed?
+  Scenario: `isa $x` matches instances that have `$x` as an inferred supertype (?)
```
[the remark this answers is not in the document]
alexjpwalker: Sounds good - done

PR #64, opened 2020-06-26
Migrate axiomatic tests to BDD

alexjpwalker:

To migrate our axiomatic Graql tests to BDD.

This PR forms part of the work towards https://github.com/graknlabs/grakn/issues/5719 .

Adds new BDD tests, many of which were migrated from the graql/query package in Grakn Core.


behaviour/graql/language/insert.feature
```diff
@@ -1073,6 +1096,87 @@ Feature: Graql Insert Query
...
+      | RF0 |
+    When get answers of graql query
+      """
+      match $x <match> isa <attr>; get;
+      """
+    Then uniquely identify answer concepts
+      | x   |
+      | RF0 |
+
+    Examples:
```
flyingsilverfin, line 1124: very good

behaviour/graql/language/insert.feature
```diff
@@ -1073,6 +1096,87 @@ Feature: Graql Insert Query
...
+      | double   | length     | 2019-12-26   |
+      | boolean  | is-alive   | 3            |
+      | boolean  | is-alive   | -17.9        |
+      | boolean  | is-alive   | 2019-12-26   |
+      | datetime | start-date | 1992         |
+      | datetime | start-date | 3.14         |
+      | datetime | start-date | false        |
+      | datetime | start-date | "2019-12-26" |
+    @ignore
+    # TODO: re-enable when only true and false are accepted as boolean values
```
flyingsilverfin, line 1172: make sure this is noted down with an issue in Github and reference this test to be unignored once the issue is implemented so we don't lose track :)

behaviour/graql/language/insert.feature
```diff
@@ -1474,6 +1578,76 @@ Feature: Graql Insert Query
       | BRO |
 
 
+  Scenario: inserting a new instance type on an existing instance is fine if it is already of that type
```
flyingsilverfin, line 1581: this title doesn't read super clearly

behaviour/graql/language/insert.feature
```diff
@@ -1474,6 +1578,76 @@ Feature: Graql Insert Query
...
+      """
+    Then the integrity is validated
+    When get answers of graql query
+      """
+      match $x isa person; get;
+      """
+    Then answer size is: 1
+
+
+  Scenario: inserting a new type on an existing instance throws, if the old type is not a subtype of the new one
```
flyingsilverfin, line 1603: "inserting a new type" isn't in our current Grakn vocabulary i think - i would use "re-inserting a matched instance as an unrelated type throws"

the other testss can use similar titles i think using "re-inserting"? What do you think

behaviour/graql/language/match.feature
```diff
@@ -550,6 +561,75 @@ Feature: Graql Match Clause
...
+
+  Scenario: when matching a non-existent type label to a variable from a generic `isa` query, an error is thrown
+    Then graql get throws
+      """
+      match $x isa $type; $type type polok; get;
+      """
+    Then the integrity is validated
+
+
+  Scenario: matching that the same variable is of two types is allowed
```
flyingsilverfin, line 596: ...but returns no answers (if reasoning is off :P) should we include thatbit?

behaviour/graql/language/match.feature
```diff
@@ -770,6 +914,19 @@ Feature: Graql Match Clause
       | NIN |
 
 
+  Scenario: when querying for a non-existent attribute type id, an empty result is returned
```
flyingsilverfin, line 917: have we got a test to check that retriving an attribtue by value that doesn't exist throws or is a noop?

behaviour/graql/language/match.feature
```diff
@@ -994,6 +1196,119 @@ Feature: Graql Match Clause
       | SUS |
 
 
+  Scenario: when all instances of a type own a single attribute, matches with and without `!==` partition them
```
flyingsilverfin, line 1199: hmm intent not obvious at first reading

behaviour/graql/language/match.feature
```diff
@@ -1020,6 +1335,20 @@ Feature: Graql Match Clause
       | PER |
 
 
+  Scenario: comparing unbound variables throws an error
```
flyingsilverfin, line 1338: this is "value comparing" right?

behaviour/graql/language/insert.feature
alexjpwalker: Yes - I have a note in my Discord update to do that; just didn't get round to it yesterday!

behaviour/graql/language/match.feature
alexjpwalker: Yeah this test was migrated directly from IT but it seems needlessly bulky - we can break it down.

behaviour/graql/language/insert.feature
alexjpwalker: Changed to

`# TODO: re-enable when only true and false are accepted as boolean values (issue #5803)`

behaviour/graql/language/insert.feature
alexjpwalker: Yes re-insert seems clear enough.

behaviour/graql/language/insert.feature
alexjpwalker: Updated to `Scenario: re-inserting a matched instance as an unrelated type throws an error`

behaviour/graql/language/insert.feature
alexjpwalker: Updated to `re-inserting a matched instance as an unrelated type throws an error`

behaviour/graql/language/match.feature
alexjpwalker: Ah yes of course!

alexjpwalker: I think this was one of the first cases I wrote in this category, before I'd settled my mind on the pattern: "when blah, [an error is thrown | an empty result is returned]"

Updated to `when matching that the same variable is of two types, an empty result is returned`

behaviour/graql/language/match.feature
alexjpwalker: Added [Scenario Outline: when matching a `<type>` attribute by a value that doesn't exist, an empty answer is returned]

behaviour/graql/language/match.feature
alexjpwalker: Deleted redundant test (test steps fully contained in other tests)

behaviour/graql/language/match.feature
alexjpwalker: Renamed to `Scenario: value comparison of unbound variables throws an error`

PR #77, opened 2020-07-09
Uncomment resolution steps that have been fixed by changes to the test framework

alexjpwalker:

To expand our resolution tests.

We uncommented the resolution steps that previously had to be commented due to test framework bugs, which were fixed in https://github.com/graknlabs/grakn/pull/5823.

### DIFF
--- a/behaviour/graql/reasoner/resolution/value-predicate.feature
+++ b/behaviour/graql/reasoner/resolution/value-predicate.feature
@@ -68,7 +68,6 @@ Feature: Value Predicate Resolution
       """
 
 
-  # TODO: change 'then' block to "$x has is-old true" once implicit attribute variables are resolvable
   Scenario: a rule can infer an attribute ownership based on a value predicate
     Given for each session, graql define
       """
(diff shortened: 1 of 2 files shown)


alexjpwalker:
Note: This PR depends on https://github.com/graknlabs/grakn/pull/5823

behaviour/graql/reasoner/resolution/value-predicate.feature
```diff
@@ -79,8 +78,7 @@ Feature: Value Predicate Resolution
         $a > 0;
       },
       then {
-        $x has is-old $t;
-        $t true;
+        $x has is-old true;
```
flyingsilverfin, line 81: was this the format of the test originally?

flyingsilverfin:
Very Cool! Just one question then i'll approve :)

behaviour/graql/reasoner/resolution/value-predicate.feature
alexjpwalker: This was one of the tests that we had to modify in order to make it pass, and was one of the incentives for this PR

flyingsilverfin: good

PR #79, opened 2020-07-16
Migrate Concept Inequality and Variable Role ITs to resolution BDD

alexjpwalker:

To migrate our Concept Inequality and Variable Role reasoner ITs to resolution tests.

This PR forms part of the work towards https://github.com/graknlabs/grakn/issues/5721

Added tests that were migrated from the relevant IT files in grakn.

### DIFF
--- a/behaviour/graql/reasoner/resolution/concept-inequality.feature
+++ b/behaviour/graql/reasoner/resolution/concept-inequality.feature
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+# TODO: re-enable all steps in file when 3-hop transitivity is resolvable
 Feature: Concept Inequality Resolution
 
   Background: Set up keyspaces for resolution testing
--- a/behaviour/graql/reasoner/resolution/concept-inequality.feature
+++ b/behaviour/graql/reasoner/resolution/concept-inequality.feature
@@ -193,17 +193,18 @@ Feature: Concept Inequality Resolution
     Then answer size in reasoned keyspace is: 2
 #    Then materialised and reasoned keyspaces are the same size
 
-  #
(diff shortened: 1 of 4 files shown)


behaviour/graql/reasoner/resolution/concept-inequality.feature
```diff
@@ -0,0 +1,398 @@
...
+        $y has name 'c';
+        {$x has name 'a';} or {$x has name 'b';};
+      get;
+      """
+#    Then all answers are correct in reasoned keyspace
+    Then answer size in reasoned keyspace is: 2
+#    Then materialised and reasoned keyspaces are the same size
+
+  #
+  #   Tests a scenario in which the neq predicate binds free variables of two equivalent relations.
```
jmsfltchr, line 197: We can actually put this kind of comment for the whole scenario as the scenario description, which wont then need the hashes if all goes well

alexjpwalker: Updated, where applicable

behaviour/graql/reasoner/resolution/concept-inequality.feature
```diff
@@ -0,0 +1,402 @@
...
+#    Then all answers are correct in reasoned keyspace
+    Then answer size in reasoned keyspace is: 1
+    Then for graql query
+      """
+      match
+        $s isa state, has name 's2';
+        (related-state: $s) isa holds;
+      get;
+      """
+#    Then all answers are correct in reasoned keyspace
```
flyingsilverfin, line 126: we can make the intent of the tests clearer by adding a step to assert that the answer sets are the same, rather than using a Graql conjunction to do that for us

alexjpwalker: Added 'answer set is equivalent for graql query` BDD step

PR #82, opened 2020-07-28
Migrate OntologicalQueryIT to BDD

alexjpwalker:

To migrate OntologicalQueryIT to BDD, forming part of the work towards https://github.com/graknlabs/grakn/issues/5721

Added tests to `match.feature` in Graql and `schema-queries.feature` in Reasoner which were migrated from OntologicalQueryIT in grakn


behaviour/graql/language/match.feature
```diff
@@ -123,6 +123,74 @@ Feature: Graql Match Clause
       | THI |
 
 
+  Scenario: `sub` can be used to retrieve all instances of types that are subtypes of a given type
```
flyingsilverfin, line 126: On line 53 of this file you have a heading `SCHEMA CONCEPTS` -- i think this would better be `schema queries` now

behaviour/graql/reasoner/schema-queries.feature
```diff
@@ -0,0 +1,594 @@
...
+    # 3 people x 3 types of person {person,entity,thing}
+    # 6 friendships x 3 types of friendship {friendship, relation, thing}
+    # 1 name x 3 types of name {name,attribute,thing}
+    # = 9 + 18 + 3 = 30
+    Then answer size in reasoned keyspace is: 30
+    Then materialised and reasoned keyspaces are the same size
+
+
+  @ignore
+  # TODO: re-enable when match ($a, $b) isa $type returns 'thing', 'relation' as $type for inferred rels (grakn#5850)
```
flyingsilverfin, line 121: so this still fails?

flyingsilverfin: if so, is there an issue for it?

alexjpwalker: yes - see comment

behaviour/graql/language/match.feature
alexjpwalker: Updated to `schema queries`

PR #86, opened 2020-07-31
Migrate NegationIT to BDD

alexjpwalker:

To migrate NegationIT from grakn to BDD.

Added new tests in `negation.feature` and `match.feature` which were migrated from grakn.

Also, moved all Rule Validation tests from `define.feature` to `rule-validation.feature`.


behaviour/graql/language/rule-validation.feature
```diff
@@ -23,6 +23,509 @@ Feature: Graql Rule Validation
...
+      $p has nickname "Bob";
+    };
+    robert-has-nickname-bobby sub robert-has-nickname-bob,
+    when {
+      $p isa person, has name "Robert";
+    }, then {
+      $p has nickname "Bobby";
+    };
+    """
+    Then the integrity is validated
```
flyingsilverfin, line 528: loads of this stuff is not necessarily negation, so we should update the PR description to say this correspondingly

behaviour/graql/reasoner/negation.feature
```diff
@@ -95,149 +929,366 @@ Feature: Negation Resolution
...
       """
-    Then all answers are correct in reasoned keyspace
-    # Anna is retrieved because she is not a manager in her own employee-employment relation
-    Then answer size in reasoned keyspace is: 3
+    Then answer size in reasoned keyspace is: 0
+    Then answers are consistent across 5 executions in reasoned keyspace
+#    Then materialised and reasoned keyspaces are the same size
+
+
+  # TODO: re-enable all steps when fixed (currently takes too long) (#75)
```
flyingsilverfin, line 1137: can resolution testing even handle negation?

behaviour/graql/language/rule-validation.feature
alexjpwalker: Already included!

behaviour/graql/reasoner/negation.feature
alexjpwalker: Yes!

PR #88, opened 2020-08-06
Migrate Client-Java IT from Core

flyingsilverfin:
Migrate two new keyspace tests to do with deleting keyspaces while sessions or transactions are open, tests for persistence of data and writing in read transactions.

Migrate tests for Keyspaces, Transactionality with writing in read transaction, and ensuring data is not written without a commit.


behaviour/connection/keyspace.feature
```diff
@@ -178,4 +178,38 @@ Feature: Connection Keyspace
...
+    Then connection does not have keyspace:
+      | grakn   |
+    Then for each transaction, define query; throws exception containing "transaction is closed"
+      """
+      define person sub entity;
+      """
+
+  Scenario: delete a nonexistant keyspace throws an error
+    When connection delete keyspace; throws exception
+      | grakn   |
```
haikalpribadi, line 215: new line

PR #95, opened 2020-08-11
In Concept BDDs, don't commit transactions after error steps v2

alexjpwalker:

To stop attempting to commit transactions, or perform further concept methods, after a 'throws exception' BDD step. We have decided that this is undesirable behaviour of Grakn - not to mention that it doesn't work in any client (e.g: client-java) - and will be changed in 2.0.

Currently none of our clients are able to commit transactions after an error occurs, but this has in a way occurred "by accident" - Grakn itself does not kill transactions on error, but all of our clients happen to.

Rewrote a large number of BDD scenarios to ensure that transactions are committed before any step that 'throws exception' is executed, and to ensure that a new transaction is opened after every step that 'throws exception'.

Note: This PR was originally #91, but that PR died when the branch it was being merged into was deleted.

Note: We also ignored a couple of tests that currently are not applicable in 2.0.


alexjpwalker:
This PR depends on https://github.com/graknlabs/grakn-2.0/pull/14

PR #98, opened 2020-08-24
Refactor: clean up to use 2.0 build dependencies

lolski:
Update to use cleaned up assemble-maven, deployment targets, and new format of deployment properties in `deployment.bzl`.


WORKSPACE
```diff
@@ -91,19 +91,26 @@ load("@graknlabs_graql//dependencies/maven:artifacts.bzl", graknlabs_graql_artif
...
+###############################################################
+load("@graknlabs_client_java//dependencies/graknlabs:repositories.bzl", "graknlabs_grabl_tracing")
 graknlabs_grabl_tracing()
 
 load("@graknlabs_grabl_tracing//dependencies/maven:artifacts.bzl", graknlabs_grabl_tracing_artifacts = "artifacts")
 
-#######################################
-# Load @graknlabs_grakn_core_artifact #
-#######################################
+##############################
```
flyingsilverfin, line 111: we shouldn't be depending on any grakn core sources anymore

flyingsilverfin (changes requested):

WORKSPACE
lolski: corrected. it was already depending on artifact actually, i just accidentally changed the section name.

flyingsilverfin (approved):

PR #103, opened 2020-10-06
Update to new rule syntax

flyingsilverfin:
Find and replace the remaining uses of rules that were not updated to the new rule syntax of:
```
define/undefine
rule xyz: when { ... } then { ... }
```

* Replace rule define and undefines with new rule syntax

### DIFF
--- a/graql/language/rule-validation.feature
+++ b/graql/language/rule-validation.feature
@@ -43,12 +43,11 @@ Feature: Graql Rule Validation
       define
       nickname sub name;
       person owns nickname;
-      robert-has-nickname-bob sub rule,
-      when {
+      rule robert-has-nickname-bob: when {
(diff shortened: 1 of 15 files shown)


graql/language/rule-validation.feature
```diff
@@ -154,7 +125,7 @@ Feature: Graql Rule Validation
       person owns nickname;
       has-nickname-bob sub rule,
```
alexjpwalker, line 126:
```suggestion
      rule has-nickname-bob:
```

alexjpwalker (changes requested):
Suggested a change

alexjpwalker (approved):

jmsfltchr (approved):

PR #105, opened 2020-10-12
Increased test coverage for rule validation

maxbaxt:

Making sure all rule validation mechanism are covered by the BDD tests. This has been kept in rule-validation.feature. The aim is to have a thorough testing to give us confidence when we switch the reasoner engine in 2.0.

Added more tests in rule-validation.feature file, and also tidied the tests up a little by added nickname into the default define block in order to reduce repetition.

### DIFF
--- a/graql/language/rule-validation.feature
+++ b/graql/language/rule-validation.feature
@@ -647,7 +647,7 @@ Feature: Graql Rule Validation
       """
       Then the integrity is validated
 
-    Scenario: We can implement rules with a cycle as long as there is no negation.
+    Scenario: rules with cyclic inferences are allowed as long as there is no negation
       When graql define
       """
       define
--- a/graql/language/rule-validation.feature
+++ b/graql/language/rule-validation.feature
@@ -26,9 +26,13 @@ Feature: Graql Rule Validation
     Given graql define
       """
       define
-      person sub entity, plays employment:employee, owns name, owns email @key;
+      person sub entity,
+        plays employment:employee, plays scholarship:scholar,
+        owns name, owns nickname, owns email @key;
       employment sub relation, relates employee, owns start-date;
+      scholarship sub relation, relates scholar;
       name sub attribute, value string;
+      nickname sub name;
       email sub attribute, value string;
       start-date sub attribute, value datetime;
       """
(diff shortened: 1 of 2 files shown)


graql/language/rule-validation.feature
```diff
@@ -26,9 +26,13 @@ Feature: Graql Rule Validation
...
       """
       define
-      person sub entity, plays employment:employee, owns name, owns email @key;
+      person sub entity,
+        plays employment:employee, plays scholarship:scholar,
+        owns name, owns nickname, owns email @key;
       employment sub relation, relates employee, owns start-date;
+      scholarship sub relation, relates scholar;
       name sub attribute, value string;
+      nickname sub name;
```
alexjpwalker, line 35:
```suggestion
      nickname sub attribute, value string;
```
In Grakn 2.0, the parent of a sub-attribute must be marked as `abstract`.

Given that none of the current tests actually make use of the type hierarchy, let's change the definition of `nickname`.

graql/language/rule-validation.feature
```diff
@@ -577,3 +569,122 @@ Feature: Graql Rule Validation
...
+      rule scholarship-means-employment: when {
+        $person isa person;
+        (scholar: $person) isa scholarship;
+      } then {
+        (employee: $person) isa employment;
+      };
+      """
+      Then the integrity is validated
+
+    Scenario: We can implement rules with a cycle as long as there is no negation.
```
alexjpwalker, line 650: Let's stick to the same style as other scenario descriptions:

```suggestion
    Scenario: rules with cyclic inferences are allowed as long as there is no negation
```

graql/language/rule-validation.feature
```diff
@@ -391,22 +416,6 @@ Feature: Graql Rule Validation
     Then the integrity is validated
 
 
-  Scenario: when a rule infers a relation with an incorrect roleplayer, an error is thrown
```
alexjpwalker, line 394: Is this scenario intentionally deleted?

alexjpwalker (changes requested):

alexjpwalker (approved):

PR #107, opened 2020-10-14
Rule testing updated to reflect new rule conclusion restrictions

maxbaxt:
## Background

Grakn 2.0 impose the following restrictions: 
- Restriction 1: You cannot down cast or side cast in an inference
- Restriction 2: You cannot add new roles to existing relations

Implement new rules to test these restrictions are working properly, and disable all current tests that break these new restrictions.

- New rules added in rule_validation.features
- Rules that break restrictions have @ignore flag added


graql/language/rule-validation.feature
```diff
@@ -702,3 +702,60 @@ Feature: Graql Rule Validation
...
+      """
+      define
+
+      man sub person;
+      boy sub person;
+
+      rule male-children-are-boys:
+      when {
+        $p isa person;
+        $p has age $a;
```
alexjpwalker, line 737: This may throw for the wrong reason - `age` is not defined in the knowledge graph. It should be added, either in this test or in the scenario background, and `person` should own it.

graql/language/rule-validation.feature
```diff
@@ -688,7 +688,7 @@ Feature: Graql Rule Validation
 
 #CHECK CLAUSE VALIDATION
```
alexjpwalker, line 689: I'm not sure these comments (lines 689, 707, 746) add much value. We should probably just delete them.

You can add an optional scenario description below a scenario name - just any string that doesn't start with `When`, `Then`, `And` ... or any of the other Cucumber keywords. But let's only do this when necessary to clarify the intended purpose of the scenario!

Examples of this can be found in `recursion.feature`.

graql/reasoner/negation.feature
```diff
@@ -541,6 +541,8 @@ Feature: Negation Resolution
   # MATCHING INFERRED CONCEPTS #
   ##############################
 
+  #  TODO: disabled because added new roles to existing relations no longer allowed in rule heads
+  @ignore
```
alexjpwalker, line 545: Did we consciously decide to ignore these 3 tests, as opposed to just outright deleting them?

The scenarios being tested no longer make sense - they all specifically test negation against inferred attached roleplayers. If we're settled on banning roleplayer attachment in Reasoner, we should delete the scenarios. If we change our mind later, we can restore the scenarios from version control.

graql/reasoner/recursion.feature
```diff
@@ -1633,6 +1633,8 @@ Feature: Recursion Resolution
 
 
   # TODO: re-enable all steps when resolvable (currently takes too long) (#75)
+  #  TODO: disabled because down and side casting no longer allowed
+  @ignore
```
alexjpwalker, line 1637: I don't believe this scenario does any downcasting or sidecasting.

graql/reasoner/roleplayer-attachment.feature
```diff
@@ -59,6 +59,8 @@ Feature: Roleplayer Attachment Resolution
       """
```
alexjpwalker, line 59: If we're settled on banning roleplayer attachment in Reasoner, we should delete this entire file. If we change our mind later, we can restore it from version control.

graql/reasoner/rule-interaction.feature
```diff
@@ -54,62 +54,111 @@ Feature: Rule Interaction Resolution
...
+      };
+
+      rule: wives-called-tracey:
+      when {
+          $x isa person;
+          $y isa person;
+          (husband: $x, wife: $y) isa marriage;
+      } then {
+          $y has name 'tracey';
+      };
```
alexjpwalker, line 145: Is the intention of this scenario to show that when the `when` blocks are both alpha equivalent, and the `then` blocks are both alpha equivalent, but when the `when` and `then considered together are clearly non-equivalent, that Reasoner still sees the two rules as distinct, and hence resolves both of them?

graql/reasoner/rule-interaction.feature
```diff
@@ -54,62 +54,111 @@ Feature: Rule Interaction Resolution
...
+      rule husbands-called-tracey:
+      when {
+          $x isa person;
+          $y isa person;
+          (husband: $x, wife: $y) isa marriage;
+      } then {
+          $x has name 'tracey';
+      };
+
+      rule: wives-called-tracey:
```
alexjpwalker, line 138:
```suggestion
      rule wives-called-tracey:
```

graql/reasoner/schema-queries.feature
```diff
@@ -460,6 +448,8 @@ Feature: Schema Query Resolution (Variable Types)
 
 
   # TODO: re-enable all steps when type generation is supported in resolution test framework (#75)
+#  TODO: disabled because down and side casting no longer allowed
```
alexjpwalker, line 451: Same comment as above - if we're settled on banning type generation, we should delete these scenarios entirely

graql/reasoner/type-generation.feature
```diff
@@ -29,6 +29,8 @@ Feature: Type Generation Resolution
 
 
   # TODO: re-enable all steps once attribute re-attachment is resolvable
+  #  TODO: disabled because down and side casting no longer allowed
```
alexjpwalker, line 32: Same comment as above - if we're settled on banning type generation, we should delete these scenarios entirely

graql/reasoner/type-generation.feature
```diff
@@ -167,6 +173,7 @@ Feature: Type Generation Resolution
 
   @ignore
   # TODO: re-enable when grakn#5824 is fixed (matching a variable with two types)
+    #  TODO: disabled because down and side casting no longer allowed
   Scenario: additional types for entities can be derived via attributes that they own
     Given for each session, graql define
       """
```
alexjpwalker, line 179: There's a scenario later on in this file named `Scenario: additional types for relations can be derived using direct 'isa!'`. It actually isn't using type generation, so it shouldn't have been in this file in the first place. Let's rename it to `Scenario: a relation can be inferred based on a direct type` and move it to `relation-inference.feature` just under the scenario `Scenario: a rule can infer a relation based on ownership of any instance of a specific attribute type`.

Then, we can delete `type-generation.feature` entirely.

graql/reasoner/type-hierarchy.feature
```diff
@@ -509,6 +509,8 @@ Feature: Type Hierarchy Resolution
 
 
   # TODO: re-enable all steps once attribute re-attachment is resolvable
+  #  TODO: disabled because down and side casting no longer allowed
+  @ignore
```
alexjpwalker, line 513: Same comment as above - if we're settled on banning type generation, we should delete this scenario entirely

alexjpwalker (changes requested):

graql/language/rule-validation.feature
maxbaxt: Good point, added that

graql/language/rule-validation.feature
maxbaxt: Alright, put these comments into the Scenario description.

graql/reasoner/recursion.feature
maxbaxt: It does. rule-3 has `$y isa S;` as a conclusion

graql/reasoner/negation.feature
maxbaxt: Yup agreed, they've been deleted.

graql/reasoner/roleplayer-attachment.feature
maxbaxt: File deleted

graql/reasoner/schema-queries.feature
maxbaxt: deleted

graql/reasoner/type-generation.feature
maxbaxt: deleted

graql/reasoner/type-generation.feature
maxbaxt: Moved to where you suggested and deleted `type-generation.feature`

graql/reasoner/type-hierarchy.feature
maxbaxt: deleted

graql/reasoner/recursion.feature
alexjpwalker: Yep true, missed that one

graql/reasoner/schema-queries.feature
alexjpwalker: is this actually deleted?

maxbaxt: I mean though it was. Maybe rebase put it back (still trying that out instead of merge). Definitely deleted now.

graql/reasoner/rule-interaction.feature
maxbaxt: Yes that is precisely the test. In 1.8.3 this doesn't pass, as it only checks if the `when` and `then` are alpha-equivalent.

maxbaxt: Extended the scenario description to make this clearer.

alexjpwalker (approved):

flyingsilverfin (approved):

PR #110, opened 2020-10-16
Update to Graql and Concept behaviours

alexjpwalker:

To update our test specification with new Graql and Concept behaviours recently introduced in 2.0

- New scenario: Entity types can only commit keys if every instance owns a distinct key
- New scenario: Relation types must have at least one role in order to commit, unless they are abstract
- New scenario: a cyclic type hierarchy is not allowed
- New scenario: when inserting an attribute, the type and value can be specified in two individual statements
- New scenario: defining a rule is idempotent
- Fixed scenarios that do `undefine person owns email @key` incorrectly (should not use "@key")
- Suppress IntelliJ warnings about undefined steps - they are not supposed to be defined!

### DIFF
--- a/concept/rule/rule.feature
+++ b/concept/rule/rule.feature
@@ -15,4 +15,5 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+#noinspection CucumberUndefinedStep
 Feature: Concept Rule
--- a/concept/thing/attribute.feature
+++ b/concept/thing/attribute.feature
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+#noinspection CucumberUndefinedStep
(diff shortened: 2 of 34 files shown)


flyingsilverfin (approved):

jmsfltchr (approved):

alexjpwalker:

To update our test specification with new Graql and Concept behaviours recently introduced in 2.0

- New scenario: Entity types can only commit keys if every instance owns a distinct key
- New scenario: Relation types must have at least one role in order to commit, unless they are abstract
- New scenario: a cyclic type hierarchy is not allowed
- New scenario: when inserting an attribute, the type and value can be specified in two individual statements
- New scenario: defining a rule is idempotent
- Fixed scenarios that do `undefine person owns email @key` incorrectly (should not use "@key")
- Suppress IntelliJ warnings about undefined steps - they are not supposed to be defined!

PR #112, opened 2020-10-29
Modified Tests to Reflect new Has Rules

maxbaxt:

new has constraint means that certain tests now fail, this PR will fix or deleted them as necessary, and also added new tests to ensure these restrictions are enforced.

Some tests have been modified because the then clause never needed the attribute type specified. Others had to be deleted, but they have been instead put into a new file, arithmetic-equality.feature, because the use case of transferring the value of one attribute to another will be covered by the introduction of arithmetic in grakn.


graql/reasoner/arithmetic-equality.feature
```diff
@@ -0,0 +1,277 @@
...
+#
+
+
+#  NOTE
+#  This file is retained as a placeholder for future functionality
+#  The following tests can no longer be run due to restrictions on HAS constraints in then clauses.
+#  The ability to take the value of one attribute and give it to another will, however, be useful
+#  Therefore every test here will be ignored until further change.
+
+Feature: A description
```
alexjpwalker, line 25: delete
```suggestion
```

graql/reasoner/arithmetic-equality.feature
```diff
@@ -0,0 +1,277 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+
+#  NOTE
+#  This file is retained as a placeholder for future functionality
+#  The following tests can no longer be run due to restrictions on HAS constraints in then clauses.
+#  The ability to take the value of one attribute and give it to another will, however, be useful
+#  Therefore every test here will be ignored until further change.
```
alexjpwalker, line 23: Um, is arithmetic equality the right name for this behaviour?

alexjpwalker (changes requested):

graql/reasoner/arithmetic-equality.feature
maxbaxt: well it means using arithmetic to make two values equal. So the file name is speculative of how arithmetic will eventually be implemented, and will probably change. The file is really just a place to keep tests in the meantime.

alexjpwalker: Ok seems reasonable

alexjpwalker (approved):

PR #121, opened 2020-12-15
Make concept map validation steps more concise

Phillammon:

We updated our behavioural tests to have more atomicity in the test steps involving answer concepts. Additionally, assorted other changes have been made to streamline the process of creating scenarios, and a number of additional testing scenarios have been introduced.

Scenarios have been added for testing schema writes in data sessions and vice versa.
A large quantity of singeton data tables have been reformatted to give a single string instead. 
All instances of `Uniquely identify answer concepts` and related steps (full list below) have been restructured to give all required information as part of the single step, instead of having to route via `When concept identifiers are`.
Where `When concept identifiers are` not matched to a `answers contain explanation tree` step, the `When concept identifiers are` step has been removed.
`Group identifiers are` steps have similarly been culled.
The malfunctioning scenario for testing attempted writes in a read transaction has been altered to fail in the correct manner.

Full list of modified steps:
`uniquely identify answer concepts as`
`order of value concepts is`
`answer groups are` (most affected- looks for owner instead of an identified group now)
`group aggregate values are`

### DIFF
--- a/connection/database.feature
+++ b/connection/database.feature
@@ -23,13 +23,11 @@ Feature: Connection Database
     Given connection does not have any database
 
   Scenario: create one database
-    When  connection create database:
-      | alice |
-    Then  connection has database:
-      | alice |
+    When connection create database: alice
+    Then connection has database: alice
 
   Scenario: create many databases
-    When  connection create databases:
+    When connection create databases:
       | alice   |
       | bob     |
       | charlie |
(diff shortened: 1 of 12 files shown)


alexjpwalker (commented):
Let's get the steps implemented in Core before merging this one. Also have one comment.

flyingsilverfin:
was about to comment exactly the same stuff as alex!

flyingsilverfin (approved):
just have to implement this in core and if you want clients too!

alexjpwalker (approved):

PR #122, opened 2020-12-16
Test that attribute type hierarchies can be redefined

VladGan:

Test that attribute type hierarchies can be redefined

Redefining existing schema follows different logic to creating a fresh schema. There was a bug where Grakn would
falsely claim that an attribute type couldn't be set to abstract as its subtypes were not abstract, which was
only triggered when redefining existing schema. 

Reproducible test case for issue found in https://github.com/graknlabs/grakn/issues/5955

### DIFF
--- a/graql/language/define.feature
+++ b/graql/language/define.feature
@@ -2408,3 +2408,22 @@ Feature: Graql Define Query
       """
     Then transaction commits
     Then the integrity is validated
+
+
+  Scenario: attribute type hierarchies can be redefined
+
+    Redefining existing schema follows different logic to creating a fresh schema. There was a bug where Grakn would
+    falsely claim that an attribute type couldn't be set to abstract as its subtypes were not abstract, which was
+    only triggered when redefining existing schema. (#5955)
+
+    When graql define
+      """
+      define name sub attribute, abstract, value string; location-name sub name;
+      """
+    Then transaction commits
+    Then session opens transaction of type: write
+    Then graql define
+      """
+      define name sub attribute, abstract, value string; location-name sub name;
+      """
+    Then the integrity is validated
--- a/graql/language/define.feature
+++ b/graql/language/define.feature
@@ -1400,6 +1400,35 @@ Feature: Graql Define Query
       | NAL |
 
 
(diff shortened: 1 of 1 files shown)


graql/language/define.feature
```diff
@@ -2408,3 +2408,22 @@ Feature: Graql Define Query
       """
     Then transaction commits
     Then the integrity is validated
+
+
+  Scenario: attribute type hierarchies can be redefined
+
+    Redefining existing schema follows different logic to creating a fresh schema. There was a bug where Grakn would
+    falsely claim that an attribute type couldn't be set to abstract as its subtypes were not abstract, which was
+    only triggered when redefining existing schema. (#5955)
```
flyingsilverfin, line 2417: so i like the intent, but I don't think it makes sense to write the the complete history of why this test is created :P otherwise every test would be an essay!

You could cut it down to:
1. redefining schema follows different logic to creating a fresh one (good!)
2. then explain what is different that this test tests :)

I think that a better scenario name would also come from your description of point 

Don't need to link the Issue that is fixed in the BDD file, just put it in the description of the PR 

Come to think of it, a lot of this should be PR description, and the comment here should be only in the context of this test (points 1 and 2)

flyingsilverfin (changes requested):
plz see my comment

graql/language/define.feature
```diff
@@ -1400,6 +1400,35 @@ Feature: Graql Define Query
...
+      $location type location-name, sub name;
+      """
+    When concept identifiers are
+      |     | check | value         |
+      | NAM | label | name          |
+      | LOC | label | location-name |
+    Then uniquely identify answer concepts
+      | name | location |
+      | NAM  | LOC      |
+    Then the integrity is validated
```
alexjpwalker, line 1429: hi

alexjpwalker (changes requested):

alexjpwalker (commented):
looks good to me, will let joshua approve

flyingsilverfin (approved):

haikalpribadi (approved):

PR #124, opened 2020-12-18
Connection scenario corrections

Phillammon:

We updated the scenario "write data in schema session" to properly reflect intended behaviour- it should throw when committed, not on the write itself. 

Wording of the graql define step in database.feature has been corrected to the new format
An additional commit step has been added to the write data in schema session scenario, with the throw moved to that step.

### DIFF
--- a/connection/database.feature
+++ b/connection/database.feature
@@ -140,12 +140,11 @@ Feature: Connection Database
     When session opens transaction of type: write
     When connection delete database: grakn
     Then connection does not have database: grakn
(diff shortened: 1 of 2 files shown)


lriuui0x0 (approved):

Phillammon:

We updated the scenario "write data in schema session" to properly reflect intended behaviour- it should throw when committed, not on the write itself. 

Wording of the graql define step in database.feature has been corrected to the new format
An additional commit step has been added to the write data in schema session scenario, with the throw moved to that step.

PR #138, opened 2021-01-08
Make steps that override 'plays' refer to the overridden type by scoped label

alexjpwalker:

There was an inconsistency in our Behaviour tests: some referred to overridden role types by their label, others used the scoped label. We unified them to always use the scoped label.

- Make steps that use overridden role types always refer to them by scoped label
- Ignore a database test that throws in `@After`
- Unignore session tests that were fixed in client-java

### DIFF
--- a/concept/type/entitytype.feature
+++ b/concept/type/entitytype.feature
@@ -1099,7 +1099,7 @@ Feature: Concept Entity Type
     When entity(person) set plays role: parentship:child
     When put entity type: man
     When entity(man) set supertype: person
-    When entity(man) set plays role: fathership:father as parent
+    When entity(man) set plays role: fathership:father as parentship:parent
     Then entity(man) get playing roles contain:
       | fathership:father |
       | parentship:child  |
@@ -1117,7 +1117,7 @@ Feature: Concept Entity Type
     When relation(mothership) set relates role: mother as parent
     When put entity type: woman
     When entity(woman) set supertype: person
-    When entity(woman) set plays role: mothership:mother as parent
+    When entity(woman) set plays role: mothership:mother as parentship:parent
     Then entity(woman) get playing roles contain:
       | mothership:mother |
       | parentship:child  |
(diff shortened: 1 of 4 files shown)


concept/type/entitytype.feature
```diff
@@ -1099,7 +1099,7 @@ Feature: Concept Entity Type
     When entity(person) set plays role: parentship:child
     When put entity type: man
     When entity(man) set supertype: person
-    When entity(man) set plays role: fathership:father as parent
+    When entity(man) set plays role: fathership:father as parentship:parent
```
alexjpwalker, line 1102: It was inconsistent to use unscoped role labels here, and scoped role labels elsewhere.

connection/database.feature
```diff
@@ -126,13 +126,17 @@ Feature: Connection Database
     Then connection does not have any database
 
 
+  # TODO: currently throws in @After; re-enable when we are able to check if sessions are alive (see client-java#225)
```
alexjpwalker, line 129: Deleting the database causes the sessions to disappear remotely, but the client has no way of knowing this.

It boldly attempts to close all the sessions in the scenario tear-down `@After`, and this throws errors because they don't exist, causing the scenario to unavoidably fail.

Our proposed solution, listed in the linked issue, is to add a mechanism for checking if a session is still alive. We will be able to piggyback on our Pulse mechanism, which happens to do exactly this.

connection/session.feature
```diff
@@ -146,7 +146,7 @@ Feature: Connection Session
       | eve     |
       | frank   |
 
-  @ignore-client-java
+
```
alexjpwalker, line 149: https://github.com/graknlabs/client-java/pull/227 fixes these scenarios in client-java.

flyingsilverfin (approved):

flyingsilverfin:
I think you can improve your description and your title @alexjpwalker !

PR #142, opened 2021-01-14
Temporarily disable all failing Graql Define tests

alexjpwalker:

Our Graql Define tests were nearly passing in Core and Clients, with only a small handful of failures. So we temporarily marked the failing tests as 'ignored' allowing us to enable the Define tests in CI and guard against regressions.

- Ignore failing Graql Define tests, for now
- Comment out an unimplemented Scenario Outline (it does not sit well with Client Python's `behave` framework)
- Fix a Match test that was vacuous (not actually testing the thing it claimed to)

### DIFF
--- a/graql/language/define.feature
+++ b/graql/language/define.feature
@@ -983,14 +983,9 @@ Feature: Graql Define Query
 
 
   # TODO
-  Scenario Outline: a type can own a '<value_type>' attribute type as a key
-
-    Examples:
-
-  # TODO
-  Scenario Outline: a '<value_type>' attribute type is not allowed to be a key
-
-    Examples:
+#  Scenario Outline: a type can own a '<value_type>' attribute type as a key
+#
+#  Scenario Outline: a '<value_type>' attribute type is not allowed to be a key
 
 
   ##################
--- a/graql/language/define.feature
+++ b/graql/language/define.feature
@@ -392,6 +392,8 @@ Feature: Graql Define Query
       | label:pet-ownership |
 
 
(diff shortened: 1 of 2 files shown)


graql/language/define.feature
```diff
@@ -983,14 +989,9 @@ Feature: Graql Define Query
...
-
-    Examples:
-
-  # TODO
-  Scenario Outline: a '<value_type>' attribute type is not allowed to be a key
-
-    Examples:
+#  Scenario Outline: a type can own a '<value_type>' attribute type as a key
+#
+#  Scenario Outline: a '<value_type>' attribute type is not allowed to be a key
```
alexjpwalker, line 994: The reason Client Python's `behave` framework hated these scenarios (and threw a fatal error) was because of their empty `Examples` blocks. Given that it's syntactically invalid Cucumber to have a `Scenario Outline` with no `Examples` block, the only option that satisfies all our BDD frameworks is to comment them out for now.

graql/language/match.feature
```diff
@@ -1505,7 +1509,7 @@ Feature: Graql Match Query
     Given graql define
       """
       define
-     unit sub attribute, value string, owns unit, owns ref;
+      unit sub attribute, value string, owns unit, owns ref;
```
alexjpwalker, line 1512: Client Python's `behave` framework also crashed attempting to parse this string before this change.

flyingsilverfin: lol

flyingsilverfin (approved):

PR #145, opened 2021-01-18
Ignore Define test that opens parallel transactions in clients

alexjpwalker:

We were seeing intermittent test failures in Client Python and stalling in Clients Java and NodeJS in a particular `define` test that opened two transactions in parallel. The GraqlSteps BDD implementations are not really set up to handle this properly, which may be responsible for the issue. So, for now, we ignore this test in all clients.

See https://github.com/graknlabs/client-java/issues/233 for further information about this issue.

- Ignore "Scenario: Uncommitted transaction writes are not persisted" in Clients Java, NodeJS and Python

### DIFF
--- a/graql/language/define.feature
+++ b/graql/language/define.feature
@@ -2063,6 +2063,10 @@ Feature: Graql Define Query
   # TRANSACTIONALITY #
   ####################
 
+  # TODO: re-enable when it passes reliably in clients (see client-java#233)
+  @ignore-client-java
+  @ignore-client-nodejs
+  @ignore-client-python
   Scenario: uncommitted transaction writes are not persisted
     When graql define
       """


graql/language/define.feature
```diff
@@ -2063,6 +2063,10 @@ Feature: Graql Define Query
   # TRANSACTIONALITY #
   ####################
 
+  # TODO: re-enable when it passes reliably in clients (see client-java#233)
+  @ignore-client-java
+  @ignore-client-nodejs
+  @ignore-client-python
```
alexjpwalker, line 2069: We were seeing intermittent test failures in Client Python and stalling in Clients Java and NodeJS in a particular `define` test that opened two transactions in parallel. The GraqlSteps BDD implementations are not really set up to handle this properly, which may be responsible for the issue. So, for now, we ignore this test in all clients.

See https://github.com/graknlabs/client-java/issues/233 for further information about this issue.

flyingsilverfin (approved):

PR #147, opened 2021-01-19
Add scenario to test overriding roles in plays

Phillammon:

We add a test scenario covering the case in which a type plays a role overriding an inherited role.

We added the `types should be able to define roles they play with an override` test scenario to `define.feature`

### DIFF
--- a/graql/language/define.feature
+++ b/graql/language/define.feature
@@ -522,6 +522,19 @@ Feature: Graql Define Query
     Then the integrity is validated
 
 
+  Scenario: types should be able to define roles they play with an override
+    Then graql define
+      """
+        define
+        locates sub relation, relates located;
+        contractor-locates sub locates, relates contractor-located as located;
+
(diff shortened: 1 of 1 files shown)


flyingsilverfin:
I think the title and description can be improved:
"Add scenario to test plays can override role" (or something like this, a bit more descriptive than "coverage")
and the description has just a couple too many words for me to be confused and be able to parse it in a quick scan lol:
"We add a test scenario covering the case in which a type [fact it's a rel isn't interesting] plays a role overriding an inherited role."
Also don't really have to comment that its not functional in Grakn, as this is the domain of testing evaluating the tests :)

flyingsilverfin (approved):
Approving with the comment above ^^ please absorb and update

alexjpwalker (approved):

flyingsilverfin:
don't forget to add project and label :)

PR #149, opened 2021-01-20
Fix undefine feature tests

Phillammon:

Due to the large quantity of failing tests, undefine.feature was previously turned off in automation. We fixed the tests that were incorrectly specified in order to reenable it in automation.

Fixed the tests that were specified incorrectly.

### DIFF
--- a/graql/language/undefine.feature
+++ b/graql/language/undefine.feature
@@ -804,7 +804,7 @@ Feature: Graql Undefine Query
     Then the integrity is validated
 
 
-  Scenario: undefining a playable role that was not actually playable to begin with is a no-op
+  Scenario: undefining a playable role that was not actually playable to begin with throws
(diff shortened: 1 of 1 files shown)


haikalpribadi (approved):

graql/language/undefine.feature
```diff
@@ -119,13 +110,22 @@ Feature: Graql Undefine Query
       | label:person |
 
 
-  Scenario: if 'entity' is not the direct supertype of an entity, undefining 'sub entity' on it does nothing
+  Scenario: undefining sub a supertype that is not a type's direct supertype should still undefine that type
```
alexjpwalker, line 113:
```suggestion
  Scenario: undefining a type 'sub' an indirect supertype should still remove that type
```

PR #154, opened 2021-01-25
Update delete testing scenarios to reflect grakn 2.0

Phillammon:

We updated `delete.feature` such that its scenarios now properly reflect the expected behaviour of grakn 2.0.

- Changed "exception containing [string]" to "exception containing" while content of exceptions remain in flux
- Changed several instances matching value to match key instead.
- Removed scenarios asserting removing has constraints by type, replaced with a single scenario checking attemptiong to specify type at all in removing has throws.
- Changed erroneous read transaction to write transaction.
- In scenarios making use of multiple deletion queries for a single relation, consolidated into a single relation
- Added ignore to delete by matched type variable scenario, along with explanation (behaviour not presently implemented.
- Removed type specifications from has constraint deletions in the 'complex' scenarios

### DIFF
--- a/graql/language/delete.feature
+++ b/graql/language/delete.feature
@@ -915,6 +915,9 @@ Feature: Graql Delete Query
 #
 #  First, we will match '$role1' = ROLE meta role. Using this answer we will remove a single $x from $r via the 'production'.
 #  Next, we will match '$role1' = WORK role, and we delete another 'production' player. This repeats again for $role='production'.
+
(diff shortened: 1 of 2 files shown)


graql/language/delete.feature
```diff
@@ -373,6 +373,8 @@ Feature: Graql Delete Query
   # ROLEPLAYERS #
   ###############
 
+  #TODO: This is flaky
+  @ignore
```
flyingsilverfin, line 377: is this something we're addressing?

flyingsilverfin (approved):

PR #155, opened 2021-01-26
Remove unused integrity validation steps

flyingsilverfin:
Integrity validation steps were intended to check the state of the DB, after operations are committed. However, as this feature has not been implemented in 9months across the stack, and its value is still uncertain, we are remove all these steps.

* Delete all steps to do with integrity validation


alexjpwalker (approved):
This was definitely the most boring review ever.

PR #161, opened 2021-02-02
Add Update query tests

Phillammon:

We added a new query type, so we needed to add tests to cover this new query type (or at least, the parts unique to it, as match, insert and delete have fairly good coverage of it).

Added `update.feature`

### DIFF
--- a/graql/language/update.feature
+++ b/graql/language/update.feature
@@ -121,6 +121,41 @@ Feature: Graql Update Query
       insert (parent: $y, child: $x) isa parenthood;
       """
 
+
+  Scenario: Unrelated insertion
+    Given get answers of graql insert
+      """
+      insert
+      $u isa person, has name "Alex", has ref 0;
+      $v isa person, has name "Bob", has ref 1;
+      $w isa person, has name "Charlie", has ref 2;
+      $x isa person, has name "Darius", has ref 3;
+      $y isa person, has name "Evelyn", has ref 4;
+      $z isa person, has name "Farah", has ref 5;
+      """
+    Given transaction commits
+    Given connection open data session for database: grakn
+    Given session opens transaction of type: write
+    When graql update
+      """
+      match $p isa person;
+      delete $p;
+      insert $x isa entity;
+      """
+    Then transaction commits
+    When session opens transaction of type: read
+    When get answers of graql match
+      """
+      match $x isa entity;
+      """
+    Then answer size is: 6
+    When get answers of graql match
+      """
+      match $x isa person;
+      """
+    Then answer size is: 0
+
+
   Scenario: Complex migration
     Given get answers of graql insert
       """
@@ -188,3 +223,6 @@ Feature: Graql Update Query
       $p has name $n;
       """
     Then answer size is: 0
+
+
+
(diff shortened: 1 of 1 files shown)


graql/language/update.feature
```diff
@@ -0,0 +1,178 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+#noinspection CucumberUndefinedStep
+Feature: Graql Delete Query
```
jmsfltchr, line 19: `Graql Update Query`

graql/language/update.feature
```diff
@@ -0,0 +1,178 @@
...
+    Given connection close all sessions
+    Given connection open data session for database: grakn
+    Given session opens transaction of type: write
+
+
+  ##########
+  # THINGS #
+  ##########
+
+  Scenario: simple update
```
jmsfltchr, line 57: Capitalise first word

graql/language/update.feature
```diff
@@ -0,0 +1,178 @@
...
+    Given transaction commits
+
+    Given connection close all sessions
+    Given connection open data session for database: grakn
+    Given session opens transaction of type: write
+
+
+  ##########
+  # THINGS #
+  ##########
```
jmsfltchr, line 55: If this comment for consistency with elsewhere? If not it's meaningless as we're always talking things.

graql/language/update.feature
jmsfltchr: Also this scenario does more, it updates one instance without side effects on the other, so try to put that in the title.

graql/language/update.feature
```diff
@@ -0,0 +1,178 @@
...
+    When graql update; throws exception
+      """
+      match
+        $x isa person, has name "Alex";
+        $y isa person, has name "Bob";
+        $r (friend: $x) isa friendship;
+      delete
+        $r (friend: $x);
+      insert
+        $r (friend: $y);
```
jmsfltchr, line 109: Good test case, are there any other invalid states we could put ourselves into? We could take inspiration from our other existing tests.

graql/language/update.feature
```diff
@@ -0,0 +1,178 @@
...
+      """
+      match
+        $p isa person, has name $n;
+        $nc isa nameclass, has name $n;
+      delete
+        $p has $n;
+      insert
+        (named: $p, name: $nc) isa naming;
+      """
+    Then transaction commits
```
jmsfltchr, line 178: Isn't there a value to doing the match and checking the answers? Otherwise how do we know this was successful?

jmsfltchr (changes requested):

haikalpribadi (changes requested):
I'm leaving the detailed review of the scenarios to @jmsfltchr, so I'm just commenting on the overall formatting: if there are multiple lines of `match/delete/insert` variables, put them on a new line but at the same indentation, not deeper. If not multiple lines keep them on the same line as `match/delete/insert` keyword.

jmsfltchr (approved):

haikalpribadi (approved):

PR #171, opened 2021-02-04
Add variable roles match query

flyingsilverfin:
To make sure we cover the  original query that caused the failure in https://github.com/graknlabs/grakn/pull/6134, we add a test for a relation with a variable role and a fixed role, in a match query.

* Add test case that tripped in the traversal engine, in a specific query plan where the starting point was the role player without the variable role `$x`
* slip in a cheeky test case for anonymous attributes with predicates, in `resolution/attribute-attachment.feature`

### DIFF
--- a/graql/reasoner/concept-inequality.feature
+++ b/graql/reasoner/concept-inequality.feature
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-@ignore # TODO enable when rules can resolve negation, and we're not pathetically slow with `not { $x is $y };`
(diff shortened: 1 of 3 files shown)


graql/reasoner/concept-inequality.feature
```diff
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-@ignore # TODO enable when rules can resolve negation, and we're not pathetically slow with `not { $x is $y };`
```
alexjpwalker, line 18: lol

alexjpwalker (approved):

jmsfltchr (approved):

PR #177, opened 2021-02-16
Fix ignore tags broken by comments

erikedin:

Some Scenarios have ignore tags followed by a malformed comment. The Cucumber Gherkin reference states:
"Comments are only permitted at the start of a new line, anywhere in the feature file."
Cucumber does not give the below scenarios the ignore tag, as is presumably intended, and therefore those scenarios may be unintentionally executed.

Move each comment to the preceding line.

### DIFF
--- a/graql/language/delete.feature
+++ b/graql/language/delete.feature
@@ -1134,7 +1134,8 @@ Feature: Graql Delete Query
   # COMPLEX PATTERNS #
   ####################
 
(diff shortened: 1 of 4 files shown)


flyingsilverfin (approved):

alexjpwalker (approved):

jmsfltchr (approved):

jmsfltchr:
Thanks for the PR @erikedin!

PR #182, opened 2021-03-03
Fix unsatisfiable reasoning query

flyingsilverfin:
Fix unsatisfiable concept inequality test

* Fix test

### DIFF
--- a/graql/reasoner/concept-inequality.feature
+++ b/graql/reasoner/concept-inequality.feature
@@ -509,11 +509,6 @@ Feature: Concept Inequality Resolution
     Then answer size in reasoned database is: 6
     Then materialised and reasoned databases are the same size
 
-  # TODO enable when we can resolve repeated concludables
-  @ignore
-  # TODO: re-enable once grakn#5821 is fixed
-  # TODO: re-enable all steps once implicit attribute variables are resolvable
-  # TODO: migrate to concept-inequality.feature
   Scenario: inferred attribute matches can be simultaneously restricted by both concept type and attribute value
     Given connection close all sessions
     Given connection open schema sessions for databases:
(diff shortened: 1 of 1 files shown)


graql/reasoner/concept-inequality.feature
```diff
@@ -572,15 +569,12 @@ Feature: Concept Inequality Resolution
...
         $y has $unwantedValue;
-        not { $value is $unwantedValue; };
         $unwantedValue "Ocado";
-        $value isa! $type;
-        $unwantedValue isa! $type;
-        $type type base-attribute;
-        not { $type is $unwantedType; };
-        $unwantedType type string-attribute;
+        not { $value is $unwantedValue; };
       get $x, $value, $type;
```
jmsfltchr, line 578: `$type` is no longer present

graql/reasoner/concept-inequality.feature
```diff
@@ -572,15 +569,12 @@ Feature: Concept Inequality Resolution
...
         $x has $value;
         $y has $unwantedValue;
-        not { $value is $unwantedValue; };
         $unwantedValue "Ocado";
-        $value isa! $type;
-        $unwantedValue isa! $type;
-        $type type base-attribute;
-        not { $type is $unwantedType; };
-        $unwantedType type string-attribute;
+        not { $value is $unwantedValue; };
```
jmsfltchr, line 577: According to the test name we're restricting type and attribute value separately, so it follows that this should be `$value != $unwantedValue;`

jmsfltchr (changes requested):

graql/reasoner/concept-inequality.feature
flyingsilverfin: addressed

graql/reasoner/concept-inequality.feature
flyingsilverfin: interesting - this should probably be an issue in Graql

flyingsilverfin: addressed

jmsfltchr (approved):

PR #183, opened 2021-03-03
Unignore tests that should now pass

flyingsilverfin:
We unignore tests that now pas CI in Grakn Core, based on the TODOs noted by the tests themselves.

* Unignore a load of tests in `graql/language`, which were disable because of missing delete features, or missing reasoning

### DIFF
--- a/graql/language/match.feature
+++ b/graql/language/match.feature
@@ -1092,16 +1092,12 @@ Feature: Graql Match Query
     Then session transaction is open: false
 
 
-
-  # TODO: fix - query does not throw exception, but it should
-  @ignore
   Scenario: an error is thrown when matching an entity type as if it were a relation type
     Then graql match; throws exception
       """
       match ($x) isa person;
       """
     Then session transaction is open: false
-
 
 
   Scenario: an error is thrown when matching a non-existent type label as if it were a relation type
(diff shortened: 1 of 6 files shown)


graql/language/match.feature
```diff
@@ -1566,18 +1555,14 @@ Feature: Graql Match Query
     Then session transaction is open: false
 
 
-
-  # TODO: re-enable when fixed (it doesn't throw)
-  @ignore
-  Scenario: when matching by an attribute ownership, if the owner can't actually own it, an empty result is returned
+  Scenario: exception is thrown when matching by an attribute ownership, if the owner can't actually own it0
```
alexjpwalker, line 1558:
```suggestion
  Scenario: exception is thrown when matching by an attribute ownership, if the owner can't actually own it
```

graql/language/undefine.feature
```diff
@@ -1337,8 +1327,8 @@ Feature: Graql Undefine Query
       """
     Then answer size is: 0
 
-  #TODO: Reenable when rules actually do something
-  @ignore
+
+  @ignore # TODO enable when we can do reasoning in a schema write transaction
```
alexjpwalker, line 1331: Wasn't this TODO-comment-on-same-line-as-@ignore syntax the issue that we recently had a PR submitted to fix because it's not valid Gherkin syntax?

flyingsilverfin: lol good point

flyingsilverfin:
comments addressed, merging to clean up state and unblock work on core

PR #192, opened 2021-04-07
Add match scenario that fails due to traversal bug

flyingsilverfin:
To add a test for the case fixed by https://github.com/graknlabs/grakn/pull/6287  - adding a test in which some relations fail and others satisfy the query, but we should still find the satisfying one.

* Add test that should find the one satisifying relation
* note: we insert the relations separately in two transactions to force the rocks iterators to look them up in a particular order later

### DIFF
--- a/graql/language/match.feature
+++ b/graql/language/match.feature
@@ -1222,6 +1222,42 @@ Feature: Graql Match Query
     Then answer size is: 4
 
 
+  Scenario: When some relations do not satisfy the query, the correct ones are still found
+    Given graql define
+      """
+      define
+      car sub entity, plays ownership:owned;
+      person plays ownership:owner;
+      company plays ownership:owner;
+      ownership sub relation, relates owned, relates owner, owns is-insured;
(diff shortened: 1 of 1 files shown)


graql/language/match.feature
```diff
@@ -1222,6 +1222,42 @@ Feature: Graql Match Query
...
 
+  Scenario: When some relations do not satisfy the query, the correct ones are still found
+    Given graql define
+      """
+      define
+      car sub entity, plays ownership:owned;
+      person plays ownership:owner;
+      company plays ownership:owner;
+      ownership sub relation, relates owned, relates owner, owns is-insured;
+      is-insured sub attribute, value boolean;"
```
alexjpwalker, line 1233:
```suggestion
      is-insured sub attribute, value boolean;
```

alexjpwalker (changes requested):
Requested a change

alexjpwalker (approved):

PR #193, opened 2021-04-13
Add rule value conversion validation scenarios

flyingsilverfin:

Add rule value conversion validation tests for valid and in valid double to long and long to double conversions

* Add two scenarios for rule value conversions validation

### DIFF
--- a/graql/language/rule-validation.feature
+++ b/graql/language/rule-validation.feature
@@ -305,6 +305,36 @@ Feature: Graql Rule Validation
       """
 
 
+  Scenario: when a rule creates long attribute for a double attribute type, no error is thrown
+    Then graql define
+      """
+      define
+      person owns age;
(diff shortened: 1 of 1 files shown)


graql/language/rule-validation.feature
```diff
@@ -305,6 +305,36 @@ Feature: Graql Rule Validation
       """
 
 
+  Scenario: when a rule creates long attribute for a double attribute type, no error is thrown
```
alexjpwalker, line 308: I actually prefer a more positive description - at first glance, my eyes assumed it said "an error is thrown" and I had to look again a few times:
```suggestion
  Scenario: a rule can create a long attribute for a double attribute type
```

flyingsilverfin: updated a bunch

alexjpwalker (approved):

PR #197, opened 2021-05-06
Add scenario for redefining rules

flyingsilverfin:

To cover https://github.com/vaticle/typedb/issues/6311 we add a scenario that re-defines a rule and performs a query using inference via the new rule.

* Add a `define` scenario to put another rule with the same name but different content.

### DIFF
--- a/typeql/language/define.feature
+++ b/typeql/language/define.feature
@@ -1639,7 +1639,7 @@ Feature: TypeQL Define Query
       person owns nickname;
       rule people-bob:
       when {
-        $p isa person, has email "bob@gmail.com";
+        $p has email "bob@gmail.com";
       } then {
         $p has name "Bob";
       };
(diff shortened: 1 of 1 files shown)


typeql/language/define.feature
```diff
@@ -1617,8 +1617,48 @@ Feature: TypeQL Define Query
...
+    Then transaction commits
+    When session opens transaction of type: write
+    Given typeql define
+      """
+      define
+      nickname sub attribute, value string;
+      person owns nickname;
+      rule people-bob:
+      when {
+        $p isa person, has email "bob@gmail.com";
```
alexjpwalker, line 1642: Maybe we should change this, so that the assertion later on is truly verifying that the rule definition was actually updated.
```suggestion
        $p has email "bob@gmail.com";
```

alexjpwalker (approved):

PR #203, opened 2021-06-10
Update query delete block bans nnonymous thing vars

flyingsilverfin:

Add a scenario to test that anonymous `has` or relation variables are disallowed in an Update query's `delete` block.

* Add scenario testing update query's delete block cannot contain anonymous thing variables.

### DIFF
--- a/typeql/language/update.feature
+++ b/typeql/language/update.feature
@@ -50,6 +50,33 @@ Feature: TypeQL Update Query
     Given session opens transaction of type: write
 
 
+  Scenario: Deleting anonymous variables throws an exception
+    Given get answers of typeql insert
+      """
+      insert
+        $x isa person, has name "Alex", has ref 0;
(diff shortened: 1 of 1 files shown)


typeql/language/update.feature
```diff
@@ -50,6 +50,33 @@ Feature: TypeQL Update Query
...
+    Given transaction commits
+    Given session opens transaction of type: write
+    Then typeql update; throws exception
+      """
+      match
+      $x isa person, has ref 1;
+      delete $x has name "Alex";
+      insert $x has name "Bob";
+      """
+    Then typeql update; throws exception
```
alexjpwalker, line 70: Couldn't this exception occur simply because the transaction has been closed?

I think we need an additional `Given session opens transaction of type: write`

alexjpwalker (changes requested):

alexjpwalker (approved):

PR #206, opened 2021-07-02
Add steps for reasoner correctness verification and refactor

jmsfltchr:

We add the new steps defined for verifying reasoner soundness and completeness. We also drastically compress the steps used in reasoner scenarios. The steps we need are very prescriptive, so we remove the need to manage transactions manually.

- Add new step definitions
- Summarise a lot of preexisting steps into several more concise steps `Given schema`, `Given data`, `Given query`, `Verify answer size is`, `Verify answers are sound`, `Verify answers are complete`.


typeql/reasoner/attribute-attachment.feature
```diff
@@ -18,16 +18,8 @@
...
-    Given connection does not have any database
-    Given connection create database: reasoned
-    Given connection create database: materialised
-    Given connection open schema sessions for databases:
-      | reasoned     |
-      | materialised |
-    Given for each session, open transactions of type: write
-    Given for each session, typeql define
+  Background: Set up database
+    Given schema
```
flyingsilverfin, line 22: with the current naming convention, I'd expect to use these steps in ALL of our BDD, not just in the resolution steps. However, the implementation is just for resolution -- I think what we should do is call this `Given resolution schema`, `Given resolution data`, and `Given resolution query`

typeql/reasoner/recursion.feature
```diff
@@ -109,22 +93,17 @@ Feature: Recursion Resolution
...
+    Given query
       """
       match (subordinate: $x, superior: $y) isa big-location-hierarchy;
       """
-    Then all answers are correct in reasoned database
-    Then answer size in reasoned database is: 1
-    Then materialised and reasoned databases are the same size
+    Then verify answer size is: 1
+     Then verify answers are sound
+     Then verify answers are complete
```
flyingsilverfin, line 102: random extra spaces?

typeql/reasoner/recursion.feature
```diff
@@ -261,33 +223,27 @@ Feature: Recursion Resolution
...
       match (role31: $x, role32: $y) isa relation3;
       """
-    Then all answers are correct in reasoned database
     # Each of the two material relation1 instances should infer a single relation3 via 1-to-2 and 2-to-3
-    Then answer size in reasoned database is: 2
-    Then for each session, transaction closes
-    Given for each session, open transactions of type: read
-    Then for typeql query
+    Then verify answer size is: 2
+    # Then verify answers are sound  # Fails
```
flyingsilverfin, line 232: can we mark these with `# TODO -- Fails` and then make a brief issue that mentions that search string?

flyingsilverfin (changes requested):

typeql/reasoner/attribute-attachment.feature
jmsfltchr: Gone with `Given reasoner schema`, `Given reasoner data` and `Given reasoner query`

typeql/language/update.feature
```diff
@@ -67,7 +67,7 @@ Feature: TypeQL Update Query
       delete $x has name "Alex";
       insert $x has name "Bob";
       """
-    Then for each session, transaction closes
+    Then session transaction closes
```
flyingsilverfin, line 70: ok last question - why is this the only non-reasoning one that is changed?

jmsfltchr: Ah I removed this for consistency, I believe it's the only place we use `for each session` outside the transaction features.

flyingsilverfin (approved):

PR #212, opened 2021-12-30
Remove unnecessary TODO in value-predicate.feature

### LINKED ISSUE #76: Move resolution tests to negation feature where applicable
Some of our tests are labelled with "TODO: move to negation.feature".

This should be done once resolution tests are finished.

alexjpwalker:

There was a TODO message in value-predicate.feature suggesting to move a particular scenario to negation.feature. However it doesn't make sense to be moved there, because it contains rules. So we deleted the TODO.

Remove unnecessary TODO in value-predicate.feature

fixes #76

### DIFF
--- a/typeql/reasoner/value-predicate.feature
+++ b/typeql/reasoner/value-predicate.feature
@@ -525,7 +525,6 @@ Feature: Value Predicate Resolution
     Then verify answers are complete
 
 
-  # TODO: move to negation.feature
   Scenario: a negation can filter out variables by equality to another variable with a specified value
     Given reasoning schema
       """


typeql/reasoner/value-predicate.feature
```diff
@@ -525,7 +525,6 @@ Feature: Value Predicate Resolution
     Then verify answers are complete
 
 
-  # TODO: move to negation.feature
```
alexjpwalker, line 528: it doesn't make sense to be moved there, because it contains rules.

jmsfltchr (approved):

PR #213, opened 2022-01-05
Extend setting abstractness scenarios

flyingsilverfin:

We add two more cases to the scenarios for testing Type concept abstractness:
1. an abstract attribute with a subtype cannot be made concrete
2. a relation type that has no direct instances, but has a subtype with instances, can be made abstract

* add scenario to test error case for attribute (must only have non-abstract leaves)
* add scenario to test error case for relations (and in general things) where only direct instances matter when setting abstractness, not indirect instances

### DIFF
--- a/concept/type/attributetype.feature
+++ b/concept/type/attributetype.feature
@@ -143,24 +143,16 @@ Feature: Concept Attribute Type
     Then attribute(name) is abstract: true
     When transaction commits
     When session opens transaction of type: write
-    Then attribute(name) as(string) put: alice; throws exception
-    When session opens transaction of type: write
     When put attribute type: email, with value type: string
     Then attribute(email) is abstract: false
     When transaction commits
-    When session opens transaction of type: write
+    When session opens transaction of type: read
     Then attribute(name) is abstract: true
-    Then attribute(name) as(string) put: alice; throws exception
     When session opens transaction of type: write
     Then attribute(email) is abstract: false
     When attribute(email) set abstract: true
     Then attribute(email) is abstract: true
     When transaction commits
-    When session opens transaction of type: write
-    Then attribute(email) as(string) put: alice@email.com; throws exception
-    When session opens transaction of type: write
-    Then attribute(email) is abstract: true
-    Then attribute(email) as(string) put: alice@email.com; throws exception
     When session opens transaction of type: write
     When put attribute type: company-email, with value type: string
     When attribute(company-email) set supertype: email
--- a/concept/type/attributetype.feature
+++ b/concept/type/attributetype.feature
@@ -161,6 +161,10 @@ Feature: Concept Attribute Type
     When session opens transaction of type: write
     Then attribute(email) is abstract: true
     Then attribute(email) as(string) put: alice@email.com; throws exception
+    When session opens transaction of type: write
+    When put attribute type: company-email, with value type: string
+    When attribute(company-email) set supertype: email
+    Then attribute(email) set abstract: false; throws exception
 
   Scenario: Attribute types can be subtypes of other attribute types
     When put attribute type: first-name, with value type: string
--- a/concept/type/relationtype.feature
+++ b/concept/type/relationtype.feature
@@ -213,6 +213,29 @@ Feature: Concept Relation Type and Role Type
     Then relation(parentship) get role(parent) is abstract: true
     Then relation(parentship) get role(child) is abstract: true
     Then relation(parentship) create new instance; throws exception
+    When session opens transaction of type: write
+    When relation(parentship) set abstract: false
+    When put relation type: fathership
+    When relation(fathership) set supertype: parentship
+    When relation(fathership) set relates role: father as parent
+    When relation(fathership) set relates role: father-child as child
+    When put entity type: person
+    When entity(person) set plays role: fathership:father
+    Then transaction commits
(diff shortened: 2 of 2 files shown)


concept/type/attributetype.feature
```diff
@@ -161,6 +161,10 @@ Feature: Concept Attribute Type
     When session opens transaction of type: write
     Then attribute(email) is abstract: true
     Then attribute(email) as(string) put: alice@email.com; throws exception
+    When session opens transaction of type: write
+    When put attribute type: company-email, with value type: string
+    When attribute(company-email) set supertype: email
+    Then attribute(email) set abstract: false; throws exception
```
flyingsilverfin, line 167: we can create a concrete child attribute type of an abstract parent attribute type, but can no longer change the parent to be non-abstract

concept/type/relationtype.feature
```diff
@@ -213,6 +213,29 @@ Feature: Concept Relation Type and Role Type
...
+    When relation(fathership) set relates role: father-child as child
+    When put entity type: person
+    When entity(person) set plays role: fathership:father
+    Then transaction commits
+    When connection close all sessions
+    When connection open data session for database: typedb
+    When session opens transaction of type: write
+    Then $m = relation(fathership) create new instance
+    When $a = entity(person) create new instance
+    When relation $m add player for role(father): $a
```
flyingsilverfin, line 230: we create a relation subtype and insert an instance

concept/type/relationtype.feature
```diff
@@ -213,6 +213,29 @@ Feature: Concept Relation Type and Role Type
...
+    Then transaction commits
+    When connection close all sessions
+    When connection open schema session for database: typedb
+    When session opens transaction of type: write
+    Then relation(parentship) set abstract: true
+    Then transaction commits
+    When session opens transaction of type: read
+    Then relation(parentship) is abstract: true
```
flyingsilverfin, line 238: we confirm that we can freely change the parent to be abstract, since it still has no explicit instances

concept/type/relationtype.feature
alexjpwalker: I feel like this scenario is getting very long and it's hard to figure out what it's actually testing because it's a very long block of text.

And it feels like some of the lines are superfluous already, for example `Then relation(parentship) create new instance; throws exception` on line 215, is of course going to throw because we opened a `READ` transaction, that is surely out of scope of this scenario.

Maybe we can clean it up a little and perhaps break it down?

concept/type/attributetype.feature
```diff
@@ -161,6 +161,10 @@ Feature: Concept Attribute Type
     When session opens transaction of type: write
```
alexjpwalker, line 161: While we're here I think we should delete lines 159 and 160, they seem really superfluous (since they are duplicated just below)
```gherkin
    When session opens transaction of type: write
    Then attribute(email) as(string) put: alice@email.com; throws exception
```

concept/type/attributetype.feature
alexjpwalker: Yeah this is fair enough, I suspect we already have a similar test for TypeQL in `define.feature` but overlap is inevitable

alexjpwalker (changes requested):

concept/type/attributetype.feature
```diff
@@ -143,24 +143,20 @@ Feature: Concept Attribute Type
     Then attribute(name) is abstract: true
     When transaction commits
     When session opens transaction of type: write
-    Then attribute(name) as(string) put: alice; throws exception
-    When session opens transaction of type: write
```
flyingsilverfin, line 147: remove cases that will throw for reasons other than what we want to test for here!

concept/type/attributetype.feature
```diff
@@ -143,24 +143,20 @@ Feature: Concept Attribute Type
...
     Then attribute(email) is abstract: true
     When transaction commits
     When session opens transaction of type: write
-    Then attribute(email) as(string) put: alice@email.com; throws exception
-    When session opens transaction of type: write
-    Then attribute(email) is abstract: true
-    Then attribute(email) as(string) put: alice@email.com; throws exception
+    When put attribute type: company-email, with value type: string
+    When attribute(company-email) set supertype: email
+    Then attribute(email) set abstract: false; throws exception
```
flyingsilverfin, line 159: new case we want to cover - setting an attribute type with child types as abstract should fail

concept/type/relationtype.feature
```diff
@@ -206,13 +202,36 @@ Feature: Concept Relation Type and Role Type
...
     When transaction commits
-    When session opens transaction of type: write
-    Then relation(parentship) create new instance; throws exception
     When session opens transaction of type: read
     Then relation(parentship) is abstract: true
     Then relation(parentship) get role(parent) is abstract: true
     Then relation(parentship) get role(child) is abstract: true
-    Then relation(parentship) create new instance; throws exception
+
+  Scenario: relation and role types can be set to abstract when a subtype has instances
```
flyingsilverfin, line 210: split a new scenario for just testing setting a type as abstract when it has indirect instances

concept/type/relationtype.feature
flyingsilverfin: done!

alexjpwalker (approved):

PR #214, opened 2022-02-23
Update IID scenario to validate IID/type contradiction throws

flyingsilverfin:

New scenario to check that contradicting IID types throw.

Add scenario that use a templated IID query for testing IID contradictions throw.

### DIFF
--- a/typeql/language/match.feature
+++ b/typeql/language/match.feature
@@ -709,7 +709,7 @@ Feature: TypeQL Match Query
       """
 
 
-  Scenario: 'iid' does not match the instance with the specified internal iid with the wrong type
+  Scenario: 'iid' for a variable of a different type throws
     Given typeql define
       """
       define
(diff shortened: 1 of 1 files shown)


typeql/language/match.feature
```diff
@@ -735,11 +735,12 @@ Feature: TypeQL Match Query
       """
       match $x isa! shop;
       """
-    Then each answer does not satisfy
+    Then templated typeql match; throws exception
```
alexjpwalker, line 738: We noted that:

- The type inference engine scrapes the type of $x from the IID, knowing that the IID encodes the type. If it didn't do so, it would inflict a performance penalty.
- However, it may seem strange to the user that a query is semantically invalid because of a specific IID. A user is likely to see an IID as just a unique random identifier, like a GUID.
- We've chosen to go ahead and keep the exception, but it would be useful to improve the exception message in the future to explain that the type information is gleaned from the IID, and conflicts with other constraints in the query.

alexjpwalker (approved):
We noted that:

- The type inference engine scrapes the type of $x from the IID, knowing that the IID encodes the type. If it didn't do so, it would inflict a performance penalty.
- However, it may seem strange to the user that a query is semantically invalid because of a specific IID. A user is likely to see an IID as just a unique random identifier, like a GUID.

We've chosen to go ahead and keep the exception, but it would be useful to improve the exception message in the future to explain that the type information is gleaned from the IID, and conflicts with other constraints in the query.

PR #220, opened 2022-05-30
Test non-deterministic results on negated relation which is both inferred and retrieved

krishnangovindraj:
A rule containing a negated relation was reported to (unsoundly) return 2 or 3 answers non-deterministically. (see [typedb issue#6500](https://github.com/vaticle/typedb/issues/6500)). The relation could be retrieved from the database or inferred using rules. This behaviour isn't shown by release 2.10.0 but can be reproduced on earlier versions.

To prevent regression, the scenario in the issue is converted to a behaviour test and added to the negation test-suite.

### DIFF
--- a/typeql/reasoner/negation.feature
+++ b/typeql/reasoner/negation.feature
@@ -1139,3 +1139,70 @@ Feature: Negation Resolution
         { $y has index "aa"; } or { $y has index "ee"; } or { $y has index "ff"; } or
         { $y has index "gg"; } or { $y has index "hh"; };
       """
+
+  Scenario: Negated relation which is both retrievable and concludable must always consider both (issue#6500)
+    # In the issue, (from: door, to:common-room) was incorrectly returned, as it contradicts a retrievable.
+    Given reasoning schema
+      """
+      define
+
+      place plays passage:from,
+            plays passage:to,
+            plays reachable:from,
+            plays reachable:to;
+
+      passage sub relation,
+          relates from,
+          relates to;
+
+      reachable sub relation,
+          relates from,
+          relates to;
+
+      rule location-hierarchy-transitivity: when {
+          (superior: $a, subordinate: $b) isa location-hierarchy;
+          (superior: $b, subordinate: $c) isa location-hierarchy;
+      } then {
+          (superior: $a, subordinate: $c) isa location-hierarchy;
+      };
+
+      rule buggy-reachable-rule:
+          when {
+              $from isa place;
+              $to isa place;
+              $common-superior isa place;
+
+              (superior: $common-superior, subordinate: $from) isa location-hierarchy;
+              (from: $common-superior, to: $to) isa passage;
+              not {$common-superior is $to;};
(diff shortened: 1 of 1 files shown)


typeql/reasoner/negation.feature
```diff
@@ -1202,3 +1202,71 @@ Feature: Negation Resolution
...
+          relates to;
+
+      rule location-hierarchy-transitivity: when {
+          (superior: $a, subordinate: $b) isa location-hierarchy;
+          (superior: $b, subordinate: $c) isa location-hierarchy;
+      } then {
+          (superior: $a, subordinate: $c) isa location-hierarchy;
+      };
+
+      rule buggy-reachable-rule:
```
jmsfltchr, line 1233: Implies this is still buggy! Let's pick another name

typeql/reasoner/negation.feature
```diff
@@ -1202,3 +1202,71 @@ Feature: Negation Resolution
       """
     Then verify answer size is: 0
 #    TODO: Add a case with non-zero answers
+
+
+  Scenario: Negated relation which is both retrievable and concludable must always consider both (issue#6500)
```
jmsfltchr, line 1207: Good title, but reading it I realised that we actually don't use `retrievable` and `concludable` as terminology in the behaviour tests (the tests largely predate this terminology). This is probably best as it disassociates the implementation from the test scenarios. We refer to `inferred relation` a lot, I'm yet to find the terminology we should use for the non-inferred case. Please take a look or improvise.

typeql/reasoner/negation.feature
```diff
@@ -1202,3 +1202,71 @@ Feature: Negation Resolution
...
+    # In the issue, (from: door, to:common-room) was incorrectly returned, as it contradicts a retrievable.
```
jmsfltchr, line 1208: You can actually do without the `#` here - Cucumber expects a description at the start of a scenario.

typeql/reasoner/negation.feature
```diff
@@ -1202,3 +1202,71 @@ Feature: Negation Resolution
...
+      """
+      insert
+      $forest isa place, has name "forest";
+      $cabin isa place, has name "cabin";
+      $common-room isa place, has name "common room";
+      $fridge isa place, has name "fridge";
+
+      (superior: $forest,       subordinate: $cabin) isa location-hierarchy;
+      (superior: $cabin,        subordinate: $common-room) isa location-hierarchy;
+      (superior: $common-room,  subordinate: $fridge) isa location-hierarchy;
```
jmsfltchr, line 1257: We don't have a convention of lining up the roleplayers like this, so pretty as it is I'd say let's skip it (it's also hard to maintain).

jmsfltchr (changes requested):
Looks good so far!

jmsfltchr (changes requested):
Were you able to think of an alternative name for the scenario?

jmsfltchr (approved):

PR #224, opened 2022-06-28
Disable memory hungry soundness check

krishnangovindraj:
This PR disables a soundness check which runs out of memory on grabl when generating explanations to verify.

Disables the soundness check for the Scenario test.
Disabling this allows the updated soundness verifier to be merged - The update is a fix required to correctly verify explanations.

## Additional information
This test was enabled soon after 2.10 was released, but I missed the step of verifying that the fix would pass tests on grabl - they pass locally because I have more memory.
They were disabled prior to 2.10 due to erratic false-positives in the soundness verifier (which is what the updated soundness verifier fixes)

### DIFF
--- a/typeql/reasoner/recursion.feature
+++ b/typeql/reasoner/recursion.feature
@@ -1031,7 +1031,7 @@ Feature: Recursion Resolution
       get $y;
       """
     Then verify answer size is: 4
-    Then verify answers are sound
+    # Then verify answers are sound     # TODO: Runs out of memory on grabl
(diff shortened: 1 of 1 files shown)


krishnangovindraj:
Ensuring it'll pass on grabl this time.
Branch pointing to this commit https://github.com/vaticle/typedb/commit/c2435d729c38f8d067f2538aad9c7f5572db486a
grabl for the commit: https://grabl.io/krishnangovindraj/typedb/c2435d729c38f8d067f2538aad9c7f5572db486a

jmsfltchr (approved):

PR #226, opened 2022-07-08
Re-enable supported reasoner completeness checks

krishnangovindraj:
This PR reenables completeness checks which were disabled due to failures. Some (if not all) of these were failing due to the forward-chaining oracle not evaluating rules according to the stratification (induced by negation). This will be fixed with [PR#6606](https://github.com/vaticle/typedb/pull/6606).
Some tests are just not supported by the current oracle and remain disabled.

Uncomments various reasoner completeness checks.

### DIFF
--- a/typeql/reasoner/negation.feature
+++ b/typeql/reasoner/negation.feature
@@ -974,7 +974,7 @@ Feature: Negation Resolution
     Then verify answer size is: 0
     Then verify answers are consistent across 5 executions
     Then verify answers are sound
-    # Then verify answers are complete  # TODO: Fails
+    Then verify answers are complete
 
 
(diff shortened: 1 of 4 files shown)


typeql/reasoner/negation.feature
```diff
@@ -1130,7 +1130,6 @@ Feature: Negation Resolution
     Then verify answer size is: 5
     Then verify answers are sound
     Then verify answers are complete
-    # TODO: Check again if we correctly mean '$y isa node' when we enable this test
```
krishnangovindraj, line 1133: I think we do, but could use an extra pair of eyes

jmsfltchr (approved):

PR #228, opened 2022-10-12
Test key uniqueness across related types

dmitrii-ubskii:

We introduce tests that verify that the key attribute defined in a supertype is unique across all its subtypes.

Add tests for scenarios related to https://github.com/vaticle/typedb/issues/6543

### DIFF
--- a/typeql/language/insert.feature
+++ b/typeql/language/insert.feature
@@ -1360,6 +1360,58 @@ Feature: TypeQL Insert Query
       $y isa person, has ref 0;
       """
 
+  Scenario: instances of an inherited key must be unique among all instances of a type and its subtypes
+    Given connection close all sessions
+    Given connection open schema session for database: typedb
+    Given session opens transaction of type: write
+    Given typeql define
+      """
+      define
+      ref sub attribute, value long;
+      base sub entity, owns ref @key;
+      derived sub base;
+      """
+    Given transaction commits
+
+    Given connection close all sessions
+    Given connection open data session for database: typedb
+    Given session opens transaction of type: write
+
+    When typeql insert
+      """
+      insert $x isa base, has ref 0;
+      """
+    Then typeql insert; throws exception
+      """
+      insert $y isa derived, has ref 0;
+      """
+
+  Scenario: instances of an inherited key must be unique among all instances of its subtypes
+    Given connection close all sessions
+    Given connection open schema session for database: typedb
+    Given session opens transaction of type: write
+    Given typeql define
+      """
+      define
+      ref sub attribute, value long;
+      base sub entity, owns ref @key;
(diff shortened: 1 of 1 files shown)


grabl:
### PR Review Checklist

Do not edit the content of this comment. The PR reviewer should simply update this comment by ticking each review item below, as they get completed.

----
**Code**

- [ ] Packages, classes, and methods have a single domain of responsibility.
- [ ] Packages, classes, and methods are grouped into cohesive and consistent domain model.
- [ ] The code is canonical and the minimum required to achieve the goal.
- [ ] Modules, libraries, and APIs are easy to use, robust (foolproof and not errorprone), and tested.
- [ ] Logic and naming has clear narrative that communicates the accurate intent and responsibility of each module (e.g. method, class, etc.).
- [ ] The code is algorithmically efficient and scalable for the whole application.

**Architecture**

- [ ] Any required refactoring is completed, and the architecture does not introduce technical debt incidentally.
- [ ] Any required build and release automations are updated and/or implemented.
- [ ] Any new components follows a consistent style with respect to the pre-existing codebase.
- [ ] The architecture intuitively reflects the application domain, and is easy to understand.
- [ ] The architecture has a well-defined hierarchy of encapsulated components.
- [ ] The architecture is extensible and scalable.

typeql/language/insert.feature
```diff
@@ -1360,6 +1360,58 @@ Feature: TypeQL Insert Query
...
+    Given session opens transaction of type: write
+
+    When typeql insert
+      """
+      insert $x isa base, has ref 0;
+      """
+    Then typeql insert; throws exception
+      """
+      insert $y isa derived, has ref 0;
+      """
```
dmitrii-ubskii, line 1387: NB: because `derived isa base`, doing it in a different order _will_ trigger an exception in the current typedb build.

flyingsilverfin: can you open another tx and add the same case in reverse though? For future regressions. So fist insert derived, then base with the same key

typeql/language/insert.feature
```diff
@@ -1360,6 +1360,70 @@ Feature: TypeQL Insert Query
...
+    Given session opens transaction of type: write
+
+    When typeql insert
+      """
+      insert $y isa derived, has ref 0;
+      """
+    Then typeql insert; throws exception
+      """
+      insert $x isa base, has ref 0;
+      """
```
dmitrii-ubskii, line 1399: Once more, with feeling. Check that the exception is raised in the reverse order, too.

flyingsilverfin: nice! technically you don't need to open a new session the second time ;)

flyingsilverfin (approved):

PR #239, opened 2023-03-24
Introduce scenarios for attribute ownership uniqueness

krishnangovindraj:

Introduce extensive tests for unique constraints on attributes, generalised behaviours of annotations, and refactored the Concept API scenarios to allow for generalised annotations.

Scenarios to specify:

1. `unique` constraint enforces an attribute is owned by at most one instance.  
2. annotations are inherited not re-declarable. However, they are specialisable (such as unique -> key) in certain valid combinations, and rejected otherwise

We verify that the define/undefine/insert/match paths to do with annotations behave as expected.

We extensively refactor the syntax of the Concept API to specify annotations in a generalised style, instead of treating `key` specially.

### DIFF
--- a/concept/thing/entity.feature
+++ b/concept/thing/entity.feature
@@ -30,7 +30,7 @@ Feature: Concept Entity
     Given put attribute type: username, with value type: string
     Given put attribute type: email, with value type: string
     Given put entity type: person
-    Given entity(person) set owns key type: username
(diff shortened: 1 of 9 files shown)


benjaminasmall:
Looking pretty solid to me. But @flyingsilverfin is the expert.

flyingsilverfin (approved):

PR #241, opened 2023-04-04
Types referenced in rules cannot be undefined

krishnangovindraj:
Add a test to ensure we cannot undefined types which are referenced in some rule.

* Adds a scenario to `undefine.feature` which attempts to undefined a type which is referenced in the `then` of a rule
* Adds a scenario to `undefine.feature` which attempts to undefined a type which is referenced in the `when` of a rule
  * if the when is a conjunction
  * in a branch, if the when is a disjunction 
  * in the negation in the when of a rule

### DIFF
--- a/typeql/language/undefine.feature
+++ b/typeql/language/undefine.feature
@@ -1362,7 +1362,57 @@ Feature: TypeQL Undefine Query
       | n                 |
       | value:name:Samuel |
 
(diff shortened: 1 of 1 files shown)


flyingsilverfin:
Please also add a test for the simpler case of undefining types mentioned in rules without disjunctions, and in negations as well.

flyingsilverfin (approved):
Much better!

PR #247, opened 2023-04-27
Add test: connected user can be retrieved

jamesreprise:

We've added a behaviour test scenario testing that the connected user is retrievable. This checks that the client provides a way to get the currently logged in user.

* Added behaviour test scenario: `connected user is retrievable`.

### DIFF
--- a/connection/user.feature
+++ b/connection/user.feature
@@ -36,13 +36,13 @@ Feature: Connection Users
     Given users delete: user
     Then users not contains: user
 
-  Scenario: users can get themselves
+  Scenario: connected user is retrievable
     Given typedb starts
     Given connection opens with authentication: admin, password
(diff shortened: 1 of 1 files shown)


connection/user.feature
```diff
@@ -36,6 +36,14 @@ Feature: Connection Users
     Given users delete: user
     Then users not contains: user
 
+  Scenario: users can get themselves
```
flyingsilverfin, line 39: I think this is not really in our 'style' grammatically? Bit informal - try calling this 'connected user is retrievable' and call the step 'Then get connected user' instead of 'user get self'?

jamesreprise: Agreed, waiting on CI.

flyingsilverfin (approved):

PR #251, opened 2023-06-21
Tests insertion and retrieval of datetime attributes are timezone-invariant

shiladitya-mukherjee:

TypeQL and ConceptAPI BDD tests for verifying that insertion and retrieval of datetime attributes are timezone-invariant.

- Added a test in `attribute.feature` in `concept` and `insert.feature` in `typeql` folders which changes timezone between insertion and read query of a datetime attribute.

### DIFF
--- a/concept/thing/attribute.feature
+++ b/concept/thing/attribute.feature
@@ -173,6 +173,22 @@ Feature: Concept Attribute
     When $x = attribute(birth-date) as(datetime) get: 1990-01-01 11:22:33
     Then attribute(birth-date) get instances contain: $x
 
+  Scenario: Attribute with value type datetime can be inserted in one timezone and retrieved in another timezone with no change in the value
+    When set time-zone is: Asia/Calcutta
+    When $x = attribute(birth-date) as(datetime) put: 2001-08-23 08:30:00
+    Then attribute $x is null: false
+    Then attribute $x has type: birth-date
+    Then attribute $x has value type: datetime
+    Then attribute $x has datetime value: 2001-08-23 08:30:00
+    When transaction commits
+    When session opens transaction of type: read
+    When set time-zone is: America/Chicago
+    When $x = attribute(birth-date) as(datetime) get: 2001-08-23 08:30:00
+    Then attribute $x is null: false
+    Then attribute $x has type: birth-date
+    Then attribute $x has value type: datetime
+    Then attribute $x has datetime value: 2001-08-23 08:30:00
+
   Scenario: Attribute with value type boolean can be deleted
     When $x = attribute(is-alive) as(boolean) put: true
     When delete attribute: $x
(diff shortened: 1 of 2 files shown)


typeql/language/insert.feature
```diff
@@ -2772,3 +2770,33 @@ Parker";
       $x isa bird;
       $x iid V123;
       """
+
```
krishnangovindraj, line 2773: We seem to have 2 lines separating scenarios. Could you check and follow whatever the convention is?

typeql/language/insert.feature
```diff
@@ -238,8 +238,6 @@ Feature: TypeQL Insert Query
       insert $x isa thing;
       """
 
-
-
```
krishnangovindraj, line 242: Did you mean to do this?
We seem to keep two 2 lines between scenarios and section headers. Was this a fix for that or does it break it?

concept/thing/attribute.feature
```diff
@@ -308,6 +308,19 @@ Feature: Concept Attribute
...
+    When transaction commits
+    When session opens transaction of type: read
+    When set time-zone is: America/Chicago
+    When $x = attribute(birth-date) as(datetime) get: 2001-08-23 08:30:00
+    Then attribute $x is null: false
+    Then attribute $x has type: birth-date
+    Then attribute $x has value type: datetime
+    Then attribute $x has datetime value: 2001-08-23 08:30:00
 
-  Scenario: Attribute with value type datetime can be owned
```
krishnangovindraj, line 313: Did you mean to delete the two scenarios?  Please restore.

concept/thing/attribute.feature
```diff
@@ -308,6 +308,19 @@ Feature: Concept Attribute
 
   Scenario: Attribute with value type double can be owned
 
-  Scenario: Attribute with value type string can be owned
+  Scenario: Attribute with value type datetime can be inserted in one timezone and queried in another timezone with no change in the value
```
krishnangovindraj, line 311: It looks like the file is arranged with creation first, then retrieval, then deletion.
Could you find where the scenario fits in and add it there instead of at the end of the file?

typeql/language/insert.feature
```diff
@@ -2772,3 +2770,33 @@ Parker";
...
       """
+
+  Scenario: datetime object can be inserted and queried correctly (and time-zone-naively) regardless of OS timezone
+    Given set time-zone is: Asia/Calcutta
+    Given connection close all sessions
+    Given connection open schema session for database: typedb
+    Given session opens transaction of type: write
+    Given typeql define
+    """
+    define test_date sub attribute, value datetime;
```
krishnangovindraj, line 2781: `test-date`. Hyphens seem to be typeql convention.
And indenting the query string (as seen in the typeql insert step below).

krishnangovindraj (changes requested):
Test itself looks good, just conventions & arrangement

concept/thing/attribute.feature
shiladitya-mukherjee: But there was no test corresponding to this scenario

krishnangovindraj: Yes, but I see other tests which you've not removed.
So we should either remove all of them or leave all of them.

concept/thing/attribute.feature
```diff
@@ -173,6 +174,22 @@ Feature: Concept Attribute
     When $x = attribute(birth-date) as(datetime) get: 1990-01-01 11:22:33
     Then attribute(birth-date) get instances contain: $x
 
+  Scenario: Attribute with value type datetime can be inserted in one timezone and retrieved in another timezone with no change in the value
```
flyingsilverfin, line 177: Shorten: `Datetime attribute can be inserted in one timezone and retrieved in another with no change in the value`

typeql/language/insert.feature
```diff
@@ -2772,3 +2772,35 @@ Parker";
       $x isa bird;
       $x iid V123;
       """
+
+
+  Scenario: datetime object can be inserted and queried correctly (and time-zone-naively) regardless of OS timezone
```
flyingsilverfin, line 2777: If this test is basically doing the same thing as the other test, why is it named so differently?

Also, this should go under the `Attribute Ownership` section right? 'Edge Cases' is something we should do our best to stay away from as a category otherwise it'll just get filled up very quickly when people get too lazy to properly categorise the tests

cc @krishnangovindraj

flyingsilverfin (changes requested):
Please see 2 individual comments

flyingsilverfin (approved):

PR #255, opened 2023-07-12
Relation variable in rule 'then' must not be explicit

shiladitya-mukherjee:
BDD test to check relation variables used to infer new relations in then clause should be anonymous.

### DIFF
--- a/typeql/language/define.feature
+++ b/typeql/language/define.feature
@@ -2472,24 +2472,6 @@ Feature: TypeQL Define Query
       person sub green-giant;
       """
 
-  Scenario: variable declared in when cannot be redeclared in then
-    Then typeql define; throws exception
-      """
-      define
-      person sub entity;
-      parentship sub relation, relates father, relates son;
-      person plays parentship:father;
-      person plays parentship:son;
-      rule my-rule-label:
-          when {
-              $s isa person
-              $f isa person
-              $relation(father:$f, son:$s) isa parentship; $brother isa person;
-          } then {
-          $relation(father:$f, son:$brother) isa parentship;
-          };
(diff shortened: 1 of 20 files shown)


typeql/language/define.feature
```diff
@@ -2472,6 +2472,23 @@ Feature: TypeQL Define Query
...
+      person plays parentship:son;
+      rule my-rule-label:
+          when {
+              $s isa person
+              $f isa person
+              $relation(father:$f, son:$s) isa parentship; $brother isa person;
+          } then {
+          $relation(father:$f, son:$brother) isa parentship;
+          };
+      """
```
krishnangovindraj, line 2491: This likely fits better in rule-validation.feature

krishnangovindraj (changes requested):

typeql/language/define.feature
shiladitya-mukherjee: ok!

typeql/language/rule-validation.feature
```diff
@@ -394,6 +394,25 @@ Feature: TypeQL Rule Validation
...
+          when {
+              $s isa person
+              $f isa person
+              $relation(father:$f, son:$s) isa parentship; $brother isa person;
+          } then {
+          $relation(father:$f, son:$brother) isa parentship;
+          };
+      """
+
+
```
krishnangovindraj, line 415: This still seems out of place within the file, given there's a section of rules for errors concerning variables in the when and then

krishnangovindraj (approved):
Looks good to me.

flyingsilverfin (approved):

PR #259, opened 2023-09-11
Add TypeQL Fetch scenarios and refactor Get query steps

flyingsilverfin:

We implement a new set of behaviour tests, `fetch.feature`, which outline the expected behaviours of the new 'Fetch' query type.
We also refactor existing steps to use the new "TypeQL Get" query terminology, replacing "TypeQL Match" query terminology.

* Implement 'fetch.feature', which outlines expected behaviours of Fetch queries
* Refactor existing steps to use the new "TypeQL Get" query terminology, replacing "TypeQL Match" query terminology.
* Add missing 'get;' clauses to Get queries that were missing them, as they are now mandatory
* Scenarios to do with 'get-aggregate' returning an empty answer now no longer use the "not a value" terminology, and reflect the return of empty optionals instead.

### DIFF
--- a/query/language/fetch.feature
+++ b/query/language/fetch.feature
@@ -86,6 +86,35 @@ Feature: TypeQL Fetch Query
         "p": { "root": "entity", "label": "person" }
       }]
       """
+    When get answers of typeql fetch
+      """
+      match
+      $p type friendship:friend;
+      fetch
+      $p;
+      """
+    Then fetch answers are
+      """
+      [{
+        "p": { "root": "relation:role", "label": "friendship:friend" }
+      }]
+      """
+
+  # TODO: remove this scenario when we finish deprecating 'thing' type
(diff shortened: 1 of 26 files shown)


query/language/fetch.feature
```diff
@@ -0,0 +1,561 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+#noinspection CucumberUndefinedStep
+Feature: TypeQL Fetch Query
```
flyingsilverfin, line 19: TODO: add tests for `role-type` and `root-thing` type

PR #260, opened 2023-09-15
Remove root Thing type from concept BDD tests

dmitrii-ubskii:

We remove all direct references to the root Thing type from the concept suite of tests. In particular, it is now excluded from the lists of supertypes (supertypes of any entity type are all themselves entity types, e.g.). The root thing type also no longer has any meaningful associated requests, so its associated feature file has been removed altogether.

Note: this change does not affect `typeql` tests as the behaviour of queries around `thing` is preserved.


flyingsilverfin (approved):

flyingsilverfin:
Note: do not merge

PR #264, opened 2023-10-11
Create a small representative suite of driver query tests

dmitrii-ubskii:

The [TypeDB driver](https://github.com/vaticle/typedb-driver) repository contains four (and counting!) implementations of the driver for different languages, each of which is fully tested in CI. That leads to an unsustainable number of jobs (~50) being spawned in Factory on each push to a fork, which has caused us significant issues during development as each hotfix had the potential to completely stall our progress.

As the first step to reducing that cost, we introduce a small set of query tests (as those were the ones taking the most time) to be run frequently, whereas the full suite is relegated to be run on the main vaticle repository branches only.

Drive-by: rename the `typeql` package to `query`.


typeql/language/compute.feature
dmitrii-ubskii, line 1: Empty and unused

flyingsilverfin (approved):

flyingsilverfin:
Looks fine to me! Can't see anything obvious missing

PR #266, opened 2023-10-27
Add test for variable role types traversed in either direction

krishnangovindraj:
Adds tests for variable role types, ensuring they can be traversed in either direction.

Guards against regression for vaticle/typedb#6925


flyingsilverfin (approved):

flyingsilverfin:
I've approved but 'in either direction' in the context of "behaviour testing" is hard to understand - can we make it more descriptive somehow? Like 'matching variable role types with relations in relations'

PR #278, opened 2024-02-07
Schema migration  scenarios

krishnangovindraj:
## Usage and product changes
Adds scenarios of various schema modification operations, and tests that any operation which would cause the schema to become invalid, or the data to be inconsistent with the schema is disallowed. 

## Implementation
* Adds:
  - `schema-validation.feature` : Ensures the schema is valid given various schema modifications, or an exception is thrown
  - `data-validation.feature` :  Ensures that either data remains consistent with the schema after various modifications, or an exception is thrown.
  - `migration.feature` : Realistic scenarios of schema migration, combining multiple operations from the above two.
* Updates existing tests for the relaxed "redundant redeclaration" cases


concept/validation/schema-validation.feature
```diff
@@ -0,0 +1,733 @@
...
+    When session opens transaction of type: write
+    When entity(ent1) set owns attribute type: attr11
+    Then transaction commits
+
+    When session opens transaction of type: write
+    When entity(ent1) set supertype: ent01
+    Then transaction commits; throws exception
+
+
+  Scenario: Types which are referenced in rules may not be renamed
```
krishnangovindraj, line 580: Just flagging this as an interesting error that's always been caught at commit time.

concept/validation/migration.feature
```diff
@@ -0,0 +1,153 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+#noinspection CucumberUndefinedStep
+Feature: Data validation
```
flyingsilverfin, line 19: Mis-named

concept/validation/migration.feature
```diff
@@ -0,0 +1,153 @@
...
+    Given typedb starts
+    Given connection opens with default authentication
+    Given connection has been opened
+    Given connection does not have any database
+    Given connection create database: typedb
+    Given connection open schema session for database: typedb
+    Given session opens transaction of type: write
+
+
+  Scenario: An ownership can be moved down one type, with data in place at the lower levels
```
flyingsilverfin, line 31: I feel like there are more 'user' operations that we should test if we've got these. Let's cross check w/ james?

Some ideas:

- Moving a type to a different super type, with existing owns, plays, and data in place (all in one maybe?)
  - Variation on this that might require a multi-step operation to adjust the data?
- Moving a capability up/down the hierarchy (you have the downward ones) with data in place

flyingsilverfin (approved):
Solid set of tests added!

PR #296, opened 2024-09-13
Query pipelines BDD

krishnangovindraj:
## Usage and product changes
BDD covering common cases for query pipelines

## Implementation
BDD covering common cases for query pipelines


query/language/pipelines.feature
```diff
@@ -0,0 +1,202 @@
...
+    """
+    match
+      $p isa person, has name "Alice";
+    select $p;
+    """
+    Then uniquely identify answer concepts
+      | p              |
+      | key:ref:0      |
+
+  Scenario: A match can be chained with a delete and insert to emulate an update.
```
flyingsilverfin, line 104: Consistent spacing between scenarios please, check what we have elsewhere

query/language/pipelines.feature
```diff
@@ -0,0 +1,202 @@
...
+    """
+    Then uniquely identify answer concepts
+      | p              |
+      | key:ref:0      |
+
+    Given get answers of typeql read query
+    """
+    match
+      $p isa person, has name $name;
+    sort $name desc;
```
flyingsilverfin, line 161: Maybe you should extend this with one that sorts on name, then something else? The test says "set of variables"?

query/language/pipelines.feature
```diff
@@ -0,0 +1,202 @@
...
+      | age             |
+      | attr:age:2      |
+
+    Given get answers of typeql read query
+    """
+    match
+      $p1 isa person, has age $age;
+    sort $age desc;
+    offset 2;
+    limit 2;
```
flyingsilverfin, line 197: Can you add one that chains them longer and in different order too?
```
limit
sort
select
offset
sort
limit
```

query/language/pipelines.feature
krishnangovindraj: Renamed and moved the multi-variable case below

query/language/pipelines.feature
krishnangovindraj: Did something smaller

PR #300, opened 2024-09-25
Refactor and update reduce BDD

krishnangovindraj:
## Usage and product changes
Refactor and updates reduce BDD

## Implementation
* Strips out aggregate & grouped aggregate scenarios from `get.feature` into `reduce.feature`
* Updates the tests for 3.0 syntax.
* Some tests are still waiting on traversal work & value-type implementation


query/language/reduce.feature
```diff
@@ -0,0 +1,556 @@
...
+    $x isa person, has income 100.0, has ref 0;
+    """
+    Given transaction commits
+
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match $x isa person, has income $y;
+      select $x, $y;
+      reduce $std = std($y) within ($x);
```
flyingsilverfin, line 551: Let's require `$std? = std(y) within $x`
Basically, whenever the reducer is falliable, we must use an optionally-valued variable on the left.

query/language/reduce.feature
```diff
@@ -0,0 +1,556 @@
...
+          $f isa friendship;
+        """
+    Then answer size is: 9
+    When get answers of typeql read query
+    """
+        match
+          $x isa person;
+          $y isa name;
+          $f isa friendship;
+        reduce $count = count($x);
```
flyingsilverfin, line 75: Let's add another `count` case:
```
match
  $x ..
  $y ..
  $z ..
reduce $count = count($x, $y, $z)
```
which counts the number of times the triple exists without any optional components (Christoph will add the test with actual optional variables later, but we should be able to test it with multiple vars for now)

query/language/reduce.feature
```diff
@@ -0,0 +1,556 @@
...
+    Given connection open write transaction for database: typedb
+    Given typeql write query
+    """
+        insert $x isa person, has ref 0;
+        """
+    Given transaction commits
+
+    Given connection open read transaction for database: typedb
+    Then typeql read query; fails
+    """
```
flyingsilverfin, line 313: What's with this formatting? Different to all our existing ones no?

query/language/reduce.feature
krishnangovindraj: It doesn't parse, but I've updated it.

PR #303, opened 2024-10-08
Reimagine driver tests for 3.0 

farost:
## Usage and product changes
We reimagine driver BDDs in 3.0 to concentrate on the drivers' API testing instead of recovering the server's validations for each client.

The goals are:
* reduce the amount of time spent on driver testing without significant benefits (-> reduce CI costs and mitigate CI configuration);
* make sure that all the drivers implement the required features in a similar fashion;
* make the process of developing and maintaining driver BDDs (both declarations and implementations) faster and lighter.

## Implementation
The updated driver BDDs use the updated step declarations from TypeDB 3.0 server but also introduce special steps relevant only to clients to avoid the need to reimplement complex parsing algorithms in different languages.

It's expected that all the drivers maintained by TypeDB cover the following BDD packages:
* connection (mostly for "exotic" cases of a small amount);
* driver (aims to cover **all** possible API calls that should be implemented for each driver).

### DIFF
--- a/concept/migration/data-validation.feature
+++ b/concept/migration/data-validation.feature
@@ -8,8 +8,8 @@ Feature: Data validation
   Background:
     Given typedb starts
     Given connection opens with default authentication
-    Given connection has been opened
-    Given connection does not have any database
+    Given connection is open: true
+    Given connection has 0 databases
     Given connection create database: typedb
     Given connection open schema transaction for database: typedb
 
--- a/concept/migration/function.feature
+++ b/concept/migration/function.feature
@@ -8,8 +8,8 @@ Feature: Schema validation
   Background:
     Given typedb starts
     Given connection opens with default authentication
-    Given connection has been opened
-    Given connection does not have any database
+    Given connection is open: true
+    Given connection has 0 databases
     Given connection create database: typedb
     Given connection open schema transaction for database: typedb
 
@@ -17,7 +17,7 @@ Feature: Schema validation
 
 #  TODO: Refactor to functions
 #  Scenario: Types which are referenced in rules may not be renamed
-#    Given typeql define
+#    Given typeql schema query
 #    """
 #    define
 #      rel00 sub relation, relates role00;
@@ -39,7 +39,7 @@ Feature: Schema validation
 
 #  TODO: Refactor to functions
 #  Scenario: Types which are referenced in rules may not be deleted
-#    Given typeql define
+#    Given typeql schema query
 #    """
 #    define
 #      rel00 sub relation, relates role00, relates extra_role;
@@ -75,7 +75,7 @@ Feature: Schema validation
 
 #  TODO: Refactor to functions
 #  Scenario: Rules made unsatisfiable by schema modifications are flagged at commit time
-#    Given typeql define
+#    Given typeql schema query
 #    """
 #    define
 #      rel00 sub relation, relates role00, relates extra_role;
--- a/connection/transaction.feature
+++ b/connection/transaction.feature
@@ -314,7 +314,7 @@ Feature: Connection Transaction
 #      """
 #      match entity $x;
 #      """
-#    Then answer size is: 0
+#    Then answer size: 0
 
     # TODO: Fix rollback
 #  Scenario: commit after a write transaction rollback does nothing
@@ -337,7 +337,7 @@ Feature: Connection Transaction
 #      """
 #      match $x isa person;
 #      """
-#    Then answer size is: 0
+#    Then answer size: 0
 
   @ignore-typedb
   Scenario: transaction timeouts are configurable
--- a/connection/transaction.feature
+++ b/connection/transaction.feature
@@ -8,8 +8,8 @@ Feature: Connection Transaction
   Background:
     Given typedb starts
     Given connection opens with default authentication
-    Given connection has been opened
-    Given connection does not have any database
+    Given connection is open: true
+    Given connection has 0 databases
 
   Scenario Outline: one database, one <type> transaction
     When connection create database: typedb
@@ -225,7 +225,7 @@ Feature: Connection Transaction
   Scenario Outline: write in a <type> transaction fails
     Given connection create database: typedb
     Given connection open schema transaction for database: typedb
-    Given typeql define
+    Given typeql schema query
       """
       define entity person;
       """
@@ -247,7 +247,7 @@ Feature: Connection Transaction
   Scenario Outline: <command> in a schema transaction fails
     Given connection create database: typedb
     Given connection open schema transaction for database: typedb
-    Given typeql define
+    Given typeql schema query
       """
       define entity person;
       """
@@ -261,7 +261,7 @@ Feature: Connection Transaction
   Scenario Outline: <command> in a write transaction fails
     Given connection create database: typedb
     Given connection open schema transaction for database: typedb
-    Given typeql define
+    Given typeql schema query
       """
       define entity person;
       """
@@ -281,7 +281,7 @@ Feature: Connection Transaction
   Scenario: schema modification in a write transaction fails
     Given connection create database: typedb
     Given connection open write transaction for database: typedb
-    Then typeql define; fails
+    Then typeql schema query; fails
       """
       define entity person;
       """
@@ -289,7 +289,7 @@ Feature: Connection Transaction
   Scenario: write data in a schema transaction is allowed
     Given connection create database: typedb
     Given connection open schema transaction for database: typedb
-    Then typeql define
+    Then typeql schema query
       """
       define entity person;
       """
@@ -303,7 +303,7 @@ Feature: Connection Transaction
 #  Scenario: commit after a schema transaction rollback does nothing
 #    Given connection create database: typedb
 #    Given connection open schema transaction for database: typedb
-#    When typeql define
+#    When typeql schema query
 #      """
 #      define entity person;
 #      """
@@ -320,7 +320,7 @@ Feature: Connection Transaction
 #  Scenario: commit after a write transaction rollback does nothing
 #    Given connection create database: typedb
 #    Given connection open schema transaction for database: typedb
-#    Given typeql define
+#    Given typeql schema query
 #      """
 #      define entity person;
 #      """
@@ -347,12 +347,12 @@ Feature: Connection Transaction
     Given set transaction option transaction-timeout-millis to: 10000
     When session opens transaction of type: write
     Then wait 8 seconds
-    Then typeql define
+    Then typeql schema query
       """
       define entity person;
       """
     Then wait 4 seconds
-    Then typeql define; fails
+    Then typeql schema query; fails
       """
       define entity person;
       """
--- a/driver/driver.feature
+++ b/driver/driver.feature
@@ -747,7 +747,7 @@ Feature: TypeDB Driver
     Then answer get row(0) get relation(p) get type is role type: false
 
 
-  Scenario: Driver processes attributes correctly
+  Scenario Outline: Driver processes attributes of type <value-type> correctly
     Given connection open schema transaction for database: typedb
     Given typeql schema query
       """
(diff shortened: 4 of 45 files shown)


concept/migration/data-validation.feature
```diff
@@ -8,8 +8,8 @@ Feature: Data validation
   Background:
     Given typedb starts
     Given connection opens with default authentication
-    Given connection has been opened
-    Given connection does not have any database
+    Given connection is open: true
```
farost, line 11: Can choose between `is open: true / is open: false` and `is open / is not open`. I'm not against any of the options and use them in the mix.

concept/migration/function.feature
```diff
@@ -8,16 +8,16 @@ Feature: Schema validation
...
+    Given connection has 0 databases
     Given connection create database: typedb
     Given connection open schema transaction for database: typedb
 
   # TODO: Refactor existing tests to functions and add new ones!
 
 #  TODO: Refactor to functions
 #  Scenario: Types which are referenced in rules may not be renamed
-#    Given typeql define
+#    Given typeql schema query
```
farost, line 20: We now have:
* ``typeql schema query`` (to check if it fails or not)
* ``typeql write query`` (to check if it fails or not)
* ``typeql read query`` (to check if it fails or not)
* ``get answers of typeql schema query`` (to save the answers)
* ``get answers of typeql write query`` (to save the answers)
* ``get answers of typeql read query`` (to save the answers)

farost: We could probably eliminate `get answers of ...` and save the answers all the time (if the request is successful), but I decided not to change it **that** drastically =)

concept/migration/data-validation.feature
farost: These changes will affect the server. I'll create a server PR once I get approval on this one, and I'll update the server's ref in the respective PR as well.

driver/query.feature
farost, line 1: No more `query`. Just `driver` for common steps (both core and cloud) and potentially `cloud` for cloud-specific steps.

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+  Background: Open connection / create driver, create database
+    Given typedb starts
+    Given connection is open: false
+    Given connection opens with default authentication
+    Given connection is open: true
+    Given connection has 0 databases
+    Given connection create database: typedb
+    Given connection has database: typedb
+
+    # TODO: Add error strings checks
```
farost, line 22: Decided not to worry about it at the moment, as we don't check it anywhere in 3.0. But can start adding it in the driver's feature.

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+  ##############
+
+  Scenario: Driver can close connection
+    When connection closes
+    Then connection is open: false
+
+
+  Scenario: Driver can connect after an unsuccessful connection attempt
+    When connection closes
+    When connection opens with a wrong port; fails
```
farost, line 35: Don't want to use specific credentials in the BDDs, so I ended up with some weird steps like this one: "with a wrong port, please find a wrong port by yourself". But it feels more correct.

Also, I decided to continue using "fails", because it's a general keyword for every error, exception, etc. Generally, **an operation failure**. Can also be "errors", but we use "fails" on backend, so why not.

With the error message check, it can become `fails with message: ""` or smth like that.

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+  # TODO: Check options setting and retrieval
+
+
+  # TODO: Fix on_close
+#  Scenario: Driver can schedule "transaction on close" jobs
+#    Given connection does not have database: created-after-schema
+#    Given connection does not have database: created-after-write
+#    Given connection does not have database: created-after-read
+#    When connection open schema transaction for database: typedb
+#    When schedule database creation on transaction close: created-after-schema
```
farost, line 295: Is another strange step. Wanted to make something interesting with the possibility to check its execution through a common step.

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+      """
+      define entity person owns name @key; attribute name, value string;
+      """
+    When get answers of typeql read query
+      """
+      match entity $p;
+      """
+    Then answer type is: concept rows
```
farost, line 373: Initially, I liked "is concept rows" and "is not concept rows". However, it didn't want to work with the python's bdds parser when I expected any value type (for created structs) and had "is/is not" on the right without a specific separator. So I ended up using "is:" and "is not:" everywhere.

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+  Scenario: Driver processes ok query answers correctly
+    Given connection open schema transaction for database: typedb
+    When get answers of typeql schema query
+      """
+      define entity person;
+      """
+    Then answer type is: ok
+    Then answer type is not: concept rows
+    Then answer type is not: concept trees
+    Then result is a successful ok
```
farost, line 358: Needed an alternative of "get rows" for this guy.
There are generally four steps of query answer consumption:
* Get the query answer;
* (optional) Check the query answer's type;
* Unwrap the query answer to its type;
* (optional) Collect the unwrapped answer if it's an iterator

This "result is a successful ok" is specifically "as_ok()" without failures. Not sure if I need to rename it to something else, but I'm open to actions.

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+    Then answer type is not: ok
+    Then answer type is not: concept trees
+    Then answer query type is read
+    Then answer query type is not schema
+    Then answer query type is not write
+    Then answer size: 1
+    Then answer column names are:
+      | p |
+    Then answer get row(0) get entity type(p) get label: person
+    Then answer get row(0) get entity type by index of variable(p) get label: person
```
farost, line 383: I don't want to rely on header's ordering, so a safe way (+ more scalable in the code!!!) is to just say "get it by index of this variable in the header".

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+    Then answer query type is not write
+    Then answer size: 2
+    Then answer column names are:
+      | p |
+      | n |
+    Then answer get row(0) get entity type(p) get label: person
+    Then answer get row(0) get attribute type(n) get label: name
+    Then answer get row(0) get entity type by index of variable(p) get label: person
+    Then answer get row(0) get attribute type by index of variable(n) get label: name
+    Then answer get row(1) get entity type(p) get label: person
```
farost, line 426: I actually rely on ordering here a little bit as I get a specific row. Not sure if I can even if it's that simple (just two consecutive attribute types), can additionally sort it on the implementation side.

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+    Then answer get row(0) get attribute(a) is date: <is-date>
+    Then answer get row(0) get attribute(a) is datetime: <is-datetime>
+    Then answer get row(0) get attribute(a) is datetime-tz: <is-datetime-tz>
+    Then answer get row(0) get attribute(a) is duration: <is-duration>
+    Then answer get row(0) get attribute(a) is struct: false
+    Then answer get row(0) get attribute(a) get value is: <value>
+    Then answer get row(0) get attribute(a) as <value-type> is: <value>
+    Then answer get row(0) get attribute(a) get value is not: <not-value>
+    Then answer get row(0) get attribute(a) as <value-type> is not: <not-value>
+    Examples:
```
farost, line 933: Better view it in source. Don't want to change it, the table is good as is.

concept/migration/data-validation.feature
farost: Actually, here we go https://github.com/typedb/typedb/pull/7198/files

driver/driver.feature
flyingsilverfin: hmm ok, im not so worried about testing this as a primary feature yet

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+#    Then connection has database: created-after-read
+
+
+  Scenario: Driver processes transaction commit errors correctly
+    Given connection open schema transaction for database: typedb
+    When typeql schema query
+      """
+      define attribute name;
+      """
+    Then transaction commits; fails
```
flyingsilverfin, line 340: i feel like for these it would be worth having some kind of string error message to see if the driver got the right error message ( have been bitten by this before! )

driver/driver.feature
flyingsilverfin: I'd use `Then answer unwraps as ok/concept rows/concept trees`
We don't normally inidicate `successful` in the step, when success is the default assumption - we do mark fails though.

driver/driver.feature
flyingsilverfin: Fair enough, and your whole line here composes well. Let's perhaps use commas to make it a bit more readable?
`answer get row(0), get entity type by index of variable(p), get label: person`

driver/driver.feature
flyingsilverfin: Just add a sort to the query right?

flyingsilverfin: If we have it...

driver/driver.feature
```diff
@@ -0,0 +1,1265 @@
...
+    Then answer get row(0) get attribute type(a) is double: <is-double>
+    Then answer get row(0) get attribute type(a) is decimal: <is-decimal>
+    Then answer get row(0) get attribute type(a) is string: <is-string>
+    Then answer get row(0) get attribute type(a) is date: <is-date>
+    Then answer get row(0) get attribute type(a) is datetime: <is-datetime>
+    Then answer get row(0) get attribute type(a) is datetime-tz: <is-datetime-tz>
+    Then answer get row(0) get attribute type(a) is duration: <is-duration>
+    Then answer get row(0) get attribute type(a) is struct: <is-struct>
+    Examples:
+      | value-type  | is-boolean | is-long | is-double | is-decimal | is-string | is-date | is-datetime | is-datetime-tz | is-duration | is-struct |
```
flyingsilverfin, line 723: noice

flyingsilverfin (changes requested):
Minor comments

driver/driver.feature
farost: It's something we implement for the drivers, it's a part of the API, so it should be tested. If we are not ready to give it to the users, we should just comment it out in production or remove it completely.

farost: There are no more "primary" features for the drivers to test or not to test. There is a short list of features that we want to always test.
If we want something extra for a specific driver, try to cover it by integration tests, you may also use the example test for these features. If not, use BDDs and reimplement it everywhere.

driver/driver.feature
farost: Good point, fixed.

driver/driver.feature
farost: I didn't use commas in concept API bdds and tried to remove them as much as possible. The reason is that I believe that constructions like `get smth(key) ...` are already separated because of `(key)`, and it's much easier to read for me, the stages of unwrapping are clear, and there is no redundancy. 

Don't you think so?

driver/driver.feature
farost: Good point. It works for now, we can add sorting if it starts failing, I guess, because I'm not sure about the syntax atm. It's only needed in this single test.

connection/transaction.feature
```diff
@@ -97,89 +96,91 @@ Feature: Connection Transaction
...
+    Examples:
+      | type   |
+      | read   |
+    # TODO: Fix write and schema
+#      | write  |
+#      | schema |
+
 
-    # TODO: Fix parallel execution for the server
+  # TODO: Fix multiple write and schema transactions
```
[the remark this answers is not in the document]
farost: It was a part of my potential discussion! 
If we can't have a write transaction while a schema transaction is open, it should be a separate test. And then I'd test chains of `write-read-write-read` and `schema-read-schema-read`.

driver/driver.feature
farost: Sure, I'll add error messages now.

farost: Done! Added additional errors checking, remembered them in the morning.

connection/transaction.feature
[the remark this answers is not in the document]
farost: Just tell me what combinations are expected to work and what are not, and I'll write commented out tests to be verified afterwards.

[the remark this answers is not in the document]
flyingsilverfin: 1. You can always open read txn (except during schema commit operation)
2. Write transactions can be opened and committed anytime, except when a schema transaction is open
3. A schema transaction will queue and take an exclusive lock against other schema transactions, and write transactions.

driver/driver.feature
flyingsilverfin: Then we should fix these now :)

connection/transaction.feature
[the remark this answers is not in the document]
farost: So there is no way we can test non-parallel schema transactions opening while another schema or write transaction is open from the same client? Is it expected to just hang? Is there an error after a timeout, and what's the expected length of this timeout?

driver/driver.feature
flyingsilverfin: I dont want to make more work for you, so we can keep it this way. If we have a multipart step in the future that doesn't have `()`, then i'd probably introduce commas everywhere

flyingsilverfin (approved):

connection/transaction.feature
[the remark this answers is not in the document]
farost: I'll put into our task.

driver/driver.feature
farost: Okay

PR #304, opened 2024-10-16
Add database name validation into connection tests and enhance driver tests

farost:
## Usage and product changes
* We add database name validation checks to `connection/database`
* We add offset checks for `datetime-tz` in `driver/driver`
* We add basic concurrent query result processing tests in `driver/driver` with more explciit todos for the future.

### DIFF
--- a/driver/driver.feature
+++ b/driver/driver.feature
@@ -518,14 +518,10 @@ Feature: TypeDB Driver
 
   Scenario: Driver processes query errors correctly
     Given connection open schema transaction for database: typedb
-    Then typeql schema query; fails with a message containing: "[TQL"
+    Then typeql schema query; fails
       """
       """
-    Then typeql schema query; fails with a message containing: "[TQL"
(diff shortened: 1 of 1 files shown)


connection/database.feature
```diff
@@ -11,9 +11,16 @@ Feature: Connection Database
...
-    Then connection has database: alice
+  Scenario Outline: create one database with name <name>
+    When connection create database: <name>
+    Then connection has database: <name>
+    Examples:
+      | name                                                                  |
+      | alice                                                                 |
+      | ALICE                                                                 |
+      | cAn-be_Like-that_WITH-a_pretty-looooooooooooong_name-and·even‿a·smile |
+      | 資料庫                                                                   |
```
farost, line 22: Auto-formatting does not understand Chinese symbols well in terms of the space they take, so the tables will be like this.

flyingsilverfin (approved):

PR #305, opened 2024-10-23
3.0 Driver fetch tests

farost:
## Usage and product changes
We add new behaviour tests steps to test fetch responses (concept documents) received by the TypeDB drivers.

Additionally, we refactor some of the already existing steps to align with the Java driver's implementation, and cover additional API's missed before.

### DIFF
--- a/connection/database.feature
+++ b/connection/database.feature
@@ -21,7 +21,6 @@ Feature: Connection Database
       | cAn-be_Like-that_WITH-a_pretty-looooooooooooong_name-and·even‿a·smile |
       | 資料庫                                                                   |
 
-
   Scenario: create many databases
     When connection create databases:
       | alice   |
@@ -62,6 +61,10 @@ Feature: Connection Database
     Then connection create database: !; fails
     Then connection create database: ...; fails
     Then connection create database: ·‿·; fails
+
+    # TODO: Merge it with the regular "cannot create database with an incorrect name" after fixing https://github.com/typedb/typedb-driver/issues/699
+  @ignore-typedb-driver-java
+  Scenario: cannot create database with an emoji
     Then connection create database: 😎; fails
     Then connection create database: my😎database; fails
 
@@ -149,4 +152,3 @@ Feature: Connection Database
     Then transaction commits
     Then connection delete database: typedb
     Then connection does not have database: typedb
-
--- a/driver/driver.feature
+++ b/driver/driver.feature
@@ -30,7 +30,7 @@ Feature: TypeDB Driver
 
   Scenario: Driver can connect after an unsuccessful connection attempt
     When connection closes
-    When connection opens with a wrong port; fails with a message containing: "connect error"
+    When connection opens with a wrong port; fails with a message containing: "address"
     Then connection is open: false
     When connection opens with a wrong host; fails with a message containing: "failed to lookup address information"
     Then connection is open: false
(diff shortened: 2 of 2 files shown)


driver/driver.feature
```diff
@@ -30,7 +30,7 @@ Feature: TypeDB Driver
 
   Scenario: Driver can connect after an unsuccessful connection attempt
     When connection closes
-    When connection opens with a wrong port; fails with a message containing: "connect error"
+    When connection opens with a wrong port; fails with a message containing: "address"
```
farost, line 33: The error is a little bit different for the Rust's parsing and its users, decided not to worry about it.

driver/driver.feature
```diff
@@ -376,14 +376,18 @@ Feature: TypeDB Driver
...
-    Then answer type is not: concept trees
+    Then answer type is not: concept documents
     Then answer unwraps as concept rows
     Then answer query type is: read
     Then answer query type is not: schema
     Then answer query type is not: write
     Then answer size is: 1
     Then answer column names are:
       | p |
+    Then answer get row(0) query type is: read
```
farost, line 387: `answer query type is` is to check the `QueryAnwer::getQueryType()`
`answer get row(0) query type` is to check `QueryAnswer::asConceptRows::getQueryType()` (after the consumption)
We don't have it for the documents for many drivers as they receive the string representation of jsons. Not a big issue for now.

driver/driver.feature
```diff
@@ -921,8 +1131,11 @@ Feature: TypeDB Driver
     Then answer get row(0) get variable(p) is attribute: false
 
     Then answer get row(0) get variable(p) as entity
+    Then answer get row(0) get variable(p) get label: person
+    Then answer get row(0) get instance(p) get label: person
     Then answer get row(0) get instance(p) get type get label: person
-    Then answer get row(0) get entity(p) get iid exists
+    Then answer get row(0) get entity(p) contains iid
```
farost, line 1137: Wanted to reuse `contains` as a param keyword in multiple places

driver/driver.feature
```diff
@@ -1391,7 +1657,6 @@ Feature: TypeDB Driver
       insert $dt 2023-05-01T00:00:00 Asia/Calcutta isa dt;
       """
     Then answer get row(0) get attribute(dt) get value is: 2023-05-01T00:00:00 Asia/Calcutta
-    Then answer get row(0) get attribute(dt) get value is: 2023-04-30T13:30:00 America/Chicago
```
farost, line 1394: The comparison worked in Python and Rust's implementations of `Datetime`s, but the Java's implementation decided that different timezones mean different `ZonedDateTime`s (and I'm not really against it as they have a separate class for it). No need to check it there, the main thing is that we get what we inserted.

connection/database.feature
```diff
@@ -62,6 +61,10 @@ Feature: Connection Database
     Then connection create database: !; fails
     Then connection create database: ...; fails
     Then connection create database: ·‿·; fails
+
+    # TODO: Merge it with the regular "cannot create database with an incorrect name" after fixing https://github.com/typedb/typedb-driver/issues/699
+  @ignore-typedb-driver-java
+  Scenario: cannot create database with an emoji
```
dmitrii-ubskii, line 67:
```suggestion
  Scenario: cannot create database with an emoji in its name
```

PR #308, opened 2024-10-30
Uncomment is constraint test scenarios

dmitrii-ubskii:
## Usage and product changes

We uncomment the `$x is $y` constraint test scenarios now that this feature is implemented.


krishnangovindraj (approved):
Trivially good

PR #309, opened 2024-10-31
Introduce 3.0 undefine behaviour steps. Add more steps for define, redefine, match value, match as

farost:
## Usage and product changes
We update all the existing `undefine` tests to use the new 3.0 `undefine` queries, add new steps to test new 3.0 functionality (new annotations, specific errors returned for query safety), and remove some of the excessive tests (mostly non-implemented ones) for cases already covered in concept BDDs.

Additionally, we cover the missing cases for `match` queries:
* transitive owns/plays/relates;
* value;
* as.

## Implementation
Several todos are left for `match` queries that should verify that the un/re/-definitions are successful, but are not implemented in 3.0 for now. Unique kinds of these TODOs are highlighted by comments of this PR.

### DIFF
--- a/concept/thing/attribute.feature
+++ b/concept/thing/attribute.feature
@@ -59,16 +59,17 @@ Feature: Concept Attribute
     Then attribute $x has value type: <type>
     Then attribute $x has value: <value>
     Examples:
-      | attr              | type        | value                              |
-      | is-alive          | boolean     | true                               |
-      | age               | long        | 21                                 |
-      | score             | double      | 123.456                            |
-      | name              | string      | alice                              |
-      | birth-date        | date        | 1990-01-01                         |
-      | event-datetime    | datetime    | 1990-01-01T11:22:33.123456789      |
-      | global-date       | datetime-tz | 1990-01-01T11:22:33 Asia/Kathmandu |
-      | global-date       | datetime-tz | 1990-01-01T11:22:33-0100           |
-      | schedule-interval | duration    | P1Y2M3DT4H5M6.789S                 |
+      | attr              | type        | value                                                                              |
+      | is-alive          | boolean     | true                                                                               |
+      | age               | long        | 21                                                                                 |
+      | score             | double      | 123.456                                                                            |
+      | name              | string      | alice                                                                              |
+      | name              | string      | very-long-string-with_@strangESymßo¬s¡2)*(()˚¬ª#08uj!@%@£^%*&%(*@!_++±§≥≤<>?:😎資料庫 |
+      | birth-date        | date        | 1990-01-01                                                                         |
+      | event-datetime    | datetime    | 1990-01-01T11:22:33.123456789                                                      |
+      | global-date       | datetime-tz | 1990-01-01T11:22:33 Asia/Kathmandu                                                 |
+      | global-date       | datetime-tz | 1990-01-01T11:22:33-0100                                                           |
+      | schedule-interval | duration    | P1Y2M3DT4H5M6.789S                                                                 |
 
   Scenario Outline: Attribute with value type <type> can be retrieved by its value
     When $x = attribute(<attr>) put instance with value: <value>
@@ -78,16 +79,17 @@ Feature: Concept Attribute
     When $x = attribute(<attr>) get instance with value: <value>
     Then attribute(<attr>) get instances contain: $x
     Examples:
-      | attr              | type        | value                              |
-      | is-alive          | boolean     | true                               |
-      | age               | long        | 21                                 |
-      | score             | double      | 123.456                            |
-      | name              | string      | alice                              |
-      | birth-date        | date        | 1990-01-01                         |
-      | event-datetime    | datetime    | 1990-01-01T11:22:33.123456789      |
-      | global-date       | datetime-tz | 1990-01-01 11:22:33 Asia/Kathmandu |
-      | global-date       | datetime-tz | 1990-01-01T11:22:33-0100           |
-      | schedule-interval | duration    | P1Y2M3DT4H5M6.789S                 |
+      | attr              | type        | value                                                                              |
+      | is-alive          | boolean     | true                                                                               |
+      | age               | long        | 21                                                                                 |
+      | score             | double      | 123.456                                                                            |
+      | name              | string      | alice                                                                              |
+      | name              | string      | very-long-string-with_@strangESymßo¬s¡2)*(()˚¬ª#08uj!@%@£^%*&%(*@!_++±§≥≤<>?:😎資料庫 |
+      | birth-date        | date        | 1990-01-01                                                                         |
+      | event-datetime    | datetime    | 1990-01-01T11:22:33.123456789                                                      |
+      | global-date       | datetime-tz | 1990-01-01 11:22:33 Asia/Kathmandu                                                 |
+      | global-date       | datetime-tz | 1990-01-01T11:22:33-0100                                                           |
+      | schedule-interval | duration    | P1Y2M3DT4H5M6.789S                                                                 |
 
   Scenario Outline: Attribute with value type <type> can be deleted
     When $x = attribute(<attr>) put instance with value: <value>
--- a/query/language/define.feature
+++ b/query/language/define.feature
@@ -50,6 +50,39 @@ Feature: TypeQL Define Query
       | label:dog |
 
 
+  Scenario: new entity type declaration should contain kind
+    Then typeql schema query; fails
+      """
+      define dog;
+      """
+
+
+  Scenario: new entity types can be defined in multiple steps within one query
+    When typeql schema query
+      """
+      define
+      dog owns name;
+      dog plays income:earner;
+      dog @abstract;
+      dog;
+      entity dog;
+      """
+    Then transaction commits
+
+    When connection open schema transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+      $x owns name;
+      $x plays income:earner;
+      entity $x;
+      """
+    Then uniquely identify answer concepts
+      | x            |
+      | label:person |
+      | label:dog    |
+
+
   Scenario: a new entity type can be defined as a subtype, creating a new child of its parent type
     When typeql schema query
       """
@@ -332,7 +365,7 @@ Feature: TypeQL Define Query
 #      | label:house |
 
 
-  Scenario: defining a type without a kind throws
+  Scenario: defining a type without a kind errors
     Then typeql schema query; fails
       """
       define flying-spaghetti-monster;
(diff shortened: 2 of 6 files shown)


concept/thing/attribute.feature
```diff
@@ -59,16 +59,17 @@ Feature: Concept Attribute
...
-      | event-datetime    | datetime    | 1990-01-01T11:22:33.123456789      |
-      | global-date       | datetime-tz | 1990-01-01T11:22:33 Asia/Kathmandu |
-      | global-date       | datetime-tz | 1990-01-01T11:22:33-0100           |
-      | schedule-interval | duration    | P1Y2M3DT4H5M6.789S                 |
+      | attr              | type        | value                                                                              |
+      | is-alive          | boolean     | true                                                                               |
+      | age               | long        | 21                                                                                 |
+      | score             | double      | 123.456                                                                            |
+      | name              | string      | alice                                                                              |
+      | name              | string      | very-long-string-with_@strangESymßo¬s¡2)*(()˚¬ª#08uj!@%@£^%*&%(*@!_++±§≥≤<>?:😎資料庫 |
```
farost, line 67: We didn't actually test long strings to trigger value search for string values in the server logic, so I added this case here.

query/language/define.feature
farost, line 1: Mostly added new cases to cover errors like "if already defined but different (different ordering, different sub, different specialise), error".

query/language/define.feature
```diff
@@ -200,15 +216,16 @@ Feature: TypeQL Define Query
...
-    Given connection open read transaction for database: typedb
-    When get answers of typeql read query
-      """
-      match $x owns email @key;
-      """
-    Then uniquely identify answer concepts
-      | x            |
-      | label:person |
-      | label:child  |
+    # TODO: Uncomment when "match @..." is implemented
```
farost, line 219: First kind of matches: match with annotations.

query/language/define.feature
```diff
@@ -69,7 +68,24 @@ Feature: TypeQL Define Query
...
+      """
+      define
+      person sub parent;
+      entity parent @abstract;
+      entity not-parent;
+      """
+    Given transaction commits
+
+    When connection open schema transaction for database: typedb
+    Then typeql schema query; fails with a message containing: "different 'person sub parent' is already defined"
```
farost, line 82: Started using `fails with a message containing` which I had introduced in `driver.feature` when I am interested in a reason of the fail.

query/language/define.feature
```diff
@@ -937,17 +1032,23 @@ Feature: TypeQL Define Query
       | datetime   | graduation-date   |
 
 
-  # TODO
-#  Scenario Outline: a type can own a '<value_type>' attribute type as a key
```
farost, line 941: I just removed these `type_manager`-specific cases, because it is covered in concept BDDs. I didn't clean the existing working tests as I consider it a sufficient amount of duplicated tests to ensure that we follow the same logic through typeql execution (at least until we have the generated typeql queries based on the concept api steps we discussed earlier).

query/language/define.feature
```diff
@@ -937,17 +1032,23 @@ Feature: TypeQL Define Query
...
+  # STRUCTS #
+  ###########
 
+    # TODO 3.x: Add tests for structs
 
   ###############
   # ANNOTATIONS #
   ###############
 
-  # TODO: Add tests for structs and their fields
```
farost, line 950: It's my todo, I consider that "Add tests for structs" is enough to force the developer thinking about all tests for structs.

query/language/define.feature
```diff
@@ -980,13 +1089,32 @@ Feature: TypeQL Define Query
       | values("1", "2") |
 
 
+    # Uncomment and fill when annotations with arguments appear for entity types
```
farost, line 1092: Don't put TODOs here because it's an optional step, but I want to preserve the structure of the file. Plus, it may remind us about these steps missing even if we ignore it at first.

query/language/define.feature
```diff
@@ -1911,155 +2249,97 @@ Feature: TypeQL Define Query
...
-      $location type location-name, sub super-name;
-     
+      $name label super-name;
+      $location label location-name, sub super-name;
       """
     Then uniquely identify answer concepts
       | name             | location            |
       | label:super-name | label:location-name |
 
-  # TODO: Reenable this scenario after closing https://github.com/vaticle/typeql/issues/281
```
farost, line 1922: Closed, generalised in another test

query/language/define.feature
```diff
@@ -1911,155 +2249,97 @@ Feature: TypeQL Define Query
...
 
-    When connection open read transaction for database: typedb
     When get answers of typeql read query
       """
-      match $x label person, owns name;
+      match $x label name;
       """
     Then answer size is: 1
 
+    # TODO: ConstraintBase::ValueType is not implemented
```
farost, line 2336: The second kind of TODOs for matches.

farost: We should implement it and add steps to `match.feature`.

query/language/match.feature
```diff
@@ -245,6 +245,29 @@ Feature: TypeQL Match Clause
       | label:person |
 
 
+  Scenario: 'owns' matches types that own the specified attribute type with their inheriting subtypes
```
farost, line 248: We didn't have such steps in `match`, that's why I encountered it in `define`. Three simple tests for three capabilities.

query/language/redefine.feature
farost, line 1: Same as for `define`: mostly covering missed cases with errors for "if not defined, error", "if nothing is changed, error"

query/language/redefine.feature
```diff
@@ -98,13 +97,20 @@ Feature: TypeQL Redefine Query
       """
 
 
-  Scenario: can redefine entity type's sub
+  Scenario: can redefine entity type's sub, cannot redefine unchanged
     When typeql schema query
       """
-      redefine entity child sub person;
```
farost, line 104: Kinds are not necessary, removed in some cases so as not to confuse the reader + diverse testing.

query/language/redefine.feature
```diff
@@ -174,34 +187,42 @@ Feature: TypeQL Redefine Query
...
     When typeql schema query
       """
-      redefine entity person owns name[];
+      redefine person owns name[];
       """
     Then transaction commits
 
-    When connection open read transaction for database: typedb
-    When get answers of typeql read query
+    # TODO: Uncomment when match with lists is implemented
```
farost, line 204: TODO kind number 3: lists

query/language/redefine.feature
```diff
@@ -448,34 +504,88 @@ Feature: TypeQL Redefine Query
...
+    When transaction closes
+
+    When connection open schema transaction for database: typedb
+    When typeql schema query
+      """
+      redefine part-time-employment relates part-time-role as empty-abstract-role;
+      """
+    Then transaction commits
+
+    # TODO: Uncomment when querying for specialise works. Now it returns every `$_ relates $x`.
```
farost, line 528: TODO kind number four: `$A relates $B as $C` is not implemented. It has to be done with ValueTypes (TODO kind number two) as the first priority for match finishing.

query/language/undefine.feature
```diff
@@ -353,140 +516,195 @@ Feature: TypeQL Undefine Query
...
+    Then typeql read query; fails with a message containing: "empty-set for some variable"
       """
       match $x owns start-date;
       """
-    Then answer size is: 0
 
-
-  Scenario: removing key ownerships from a super relation type also removes them from its subtypes
-    Given typeql schema query
+# TODO: match with annotations (do we really need this test in this file? Only for undefine + match purposes...)
```
farost, line 529: TODO number 1, but I'm not sure if we want to keep this test here at all. I might completely remove it in favor of concept BDDs.

query/language/undefine.feature
```diff
@@ -495,1115 +713,1557 @@ Feature: TypeQL Undefine Query
...
+  ###########
+  # STRUCTS #
+  ###########
+
+    # TODO 3.x: Add tests for structs
+
+  #############
+  # FUNCTIONS #
+  #############
+# TODO: Write new tests for functions. Consider old Rules tests to be reimplemented if applicable
```
farost, line 1346: Left old rules tests just in case. Will be happy to remove it completely.

query/language/define.feature
farost: Haven't rebased yet though, will try rebasing and running it in a bit!

farost: Nope, still not impelmented

query/language/match.feature
```diff
@@ -2362,9 +2454,9 @@ Feature: TypeQL Match Clause
       match $phrase isa favorite-phrase;
       """
     Then uniquely identify answer concepts
-      | phrase                             |
+      | phrase                        |
       | attr:favorite-phrase:你明白了吗    |
-      | attr:favorite-phrase:בוקר טוב      |
+      | attr:favorite-phrase:בוקר טוב |
```
dmitrii-ubskii, line 2459: Is IDEA doing this weird formatting around double width characters, or is this by hand? I keep having to fix tables like this.

dmitrii-ubskii (approved):

query/language/match.feature
farost: It's IDEA, I mentioned it in one of my previous PRs you reviewed. So I decided to ignore it. Sorry to hear about your struggles =(

farost: This one is exceptionally ugly, so I fixed it =)

query/language/define.feature
farost: I can't find it in the spec though!

concept/thing/attribute.feature
flyingsilverfin: Nice

farost:
@dmitrii-ubskii I added more define and match tests in this PR as everything becomes connected. Please, check the newest commits!

query/language/undefine.feature
flyingsilverfin: I think there are useful error cases here we can retain, so let's not delete them yet and then transform them into Function tests when we are there. In fact, @krishnangovindraj  please add to your list revisiting these undefine tests for but rewritten for Functions!

query/language/define.feature
```diff
@@ -394,6 +427,38 @@ Feature: TypeQL Define Query
...
+
+
+  Scenario: new relation types can be defined in multiple steps within one query
+    When typeql schema query
+      """
+      define
+      pet-ownership relates pet-owner;
+      pet-ownership relates owned-pet;
+      pet-ownership @abstract;
+      pet-ownership;
```
dmitrii-ubskii, line 444: Do we really want to commit to supporting this empty define statement?

dmitrii-ubskii (approved):

query/language/define.feature
farost: If we decide to reject it, we can rewrite the test. The worst case is, to my mind, when we are not even aware if it's possible or not, thus not covered by tests.  
If this case doesn't exist, we may not remember to write a negative-case test when we change the behavior.

PR #312, opened 2024-11-13
Introduce 3.0 fetch BDD

farost:
## Usage and product changes
We introduce new BDD tests for fetch. These tests cover all forms of `fetch` queries from the 3.0 spec (excluding lists of attributes). 

Some of the cases are commented out with explicit `TODO`s, as there are expressions or functions that are being fixed in other branches. This will require revisiting this file later to uncomment these cases.


query/language/fetch.feature
```diff
@@ -11,855 +11,2942 @@ Feature: TypeQL Fetch Query
...
+      {
+        "role type label": "friendship:friend"
+      }
       """
 
 
-  # TODO: remove this scenario when we finish deprecating 'thing' type
-  Scenario: root thing type can be fetched
-    When get answers of typeql fetch
+  # TODO: Add tests for value types when value types match/fetch is implemented
```
farost, line 174: I guess we don't want it yet?

query/language/fetch.feature
```diff
@@ -11,855 +11,2942 @@ Feature: TypeQL Fetch Query
...
+      """
+    Then answer does not contain document:
+      """
+      {
+        "name": "Bob"
+      }
+      """
+
+
+  # TODO: Uncomment when expressions allow function calling with math operations
```
farost, line 2263: @krishnangovindraj for you

query/language/fetch.feature
```diff
@@ -11,855 +11,2942 @@ Feature: TypeQL Fetch Query
...
+      | date        | 2024-09-20                                  | "2024-09-20"                                  | "2025-09-20"                                 |
+      | datetime    | 1999-02-26T12:15:05                         | "1999-02-26T12:15:05.000000000"               | "1999-02-26T12:15:05"                        |
+      | datetime    | 1999-02-26T12:15:05.000000001               | "1999-02-26T12:15:05.000000001"               | "1999-02-26T12:15:05.000000000"              |
+      | datetime-tz | 2024-09-20T16:40:05.000000001 Europe/London | "2024-09-20T16:40:05.000000001 Europe/London" | "2024-09-20T16:40:05.000000001Europe/London" |
+      | datetime-tz | 2024-09-20T16:40:05.000000001+0100          | "2024-09-20T16:40:05.000000001+01:00"         | "2024-09-20T16:40:05.000000001+0100"         |
+      | duration    | P1Y10M7DT15H44M5.00394892S                  | "P1Y10M7DT15H44M5.003948920S"                 | "P1Y10M7DT15H44M5.00394892S"                 |
+      | duration    | P66W                                        | "P462D"                                       | "P66W"                                       |
+    # TODO: Test documents and structs
```
farost, line 827: For the future

farost (commented):
Added more context behind todos

driver/driver.feature
farost, line 1: I will probably add a more complicated structure later when I switch to drivers.

dmitrii-ubskii (approved):

query/language/fetch.feature
flyingsilverfin: Yes not available yet

query/language/fetch.feature
```diff
@@ -11,855 +11,2942 @@ Feature: TypeQL Fetch Query
...
+      };
+      """
+    Then typeql read query; fails with a message containing: "multiple mappings for one key"
+      """
+      match
+      $p isa! $t, has person-name $n;
+      fetch {
+        "first": $n,
+        "first": $n,
+      };
```
flyingsilverfin, line 1225: For future reference, this is great - but I think you don't really have to write this level of granularity for tests. I would generally consider: is the sensible data structure/check going to be consistently implemented the same way in all these scenarios (eg. the keys are stored in a map and the same logic is used in all the cases)? If yes, then the tests don't need to be duplicated. This isn't _super_ comprehensive and relies on everyone coding in a good and similar way, but it is a level aim to achieve and reduces the number of tests we have to write & maintain (at risk of missing something sometimes).

query/language/fetch.feature
```diff
@@ -11,855 +11,2942 @@ Feature: TypeQL Fetch Query
...
+        "subquery": [
+          {
+            "another person name": "Alice"
+          }
+        ]
+      }
+      """
+
+
+  Scenario: non-fetch subqueries are not permitted
```
flyingsilverfin, line 1322: Similarly here - you know that all queries are either Row pipelines or Document-producing pipelines. You can probably just take 1 example of a Row-producing pipeline

flyingsilverfin (approved):
Looks good!

query/language/fetch.feature
farost: I'd aim to be granular in our tests. We are creating a database, and, where possible, it would be great to have the behavior pinned in tests, so we don't miss anything.

I agree that I may be sometimes too granular in my checks and worries, so feel free to continue correcting me, please. Although in this specific case... I spent 15 minutes debugging why my tests were not working, and then I noticed that I had two headers duplicated and squeezed into a single json document entry. I imagined that it would be frustrating for our users and, considering that we are a statically strong database, I decided to add this check, and I expect it to be constant, and this test highlights that we **do** care about this behavior. 

So I did it partially for myself. And I think that I'm correct here

query/language/fetch.feature
farost: I want to make sure that users don't get their server crashed.

query/language/fetch.feature
flyingsilverfin: Fair enough I don't mind it - just keep in mind the overall tradeoff of speed vs accuracy. We definitely want to cover these edge cases at least once.

PR #315, opened 2024-11-22
Add function BDD

krishnangovindraj:
## Usage and product changes
Adds function BDD.

## Implementation
* Implements basic compile time check BDDs for functions
* Ports the `recursion.feature` from 2.x
* Does some re-arranging that still needs porting.


query/functions/basic.feature
```diff
@@ -0,0 +1,153 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#noinspection CucumberUndefinedStep
+Feature: Relation Inference Resolution
```
flyingsilverfin, line 6: Update title

query/functions/basic.feature
```diff
@@ -0,0 +1,153 @@
...
+
+  Background: Set up database
+    Given typedb starts
+    Given connection opens with default authentication
+    Given connection is open: true
+    Given connection has 0 databases
+    Given connection create database: typedb
+
+
+  Scenario: when matching all possible pairs inferred from n concepts, the answer size is the square of n
```
flyingsilverfin, line 16: remove 'inferred'

query/functions/basic.feature
```diff
@@ -0,0 +1,153 @@
...
+      entity person,
+      owns name,
+      plays friendship:friend;
+
+      relation friendship,
+        relates friend;
+
+      attribute name value string;
+
+      fun friends-of($who: person) -> { person } :
```
flyingsilverfin, line 31: 'people-pairs-with' (no friendships here? You can delete the relation type too)

query/functions/basic.feature
```diff
@@ -0,0 +1,153 @@
...
+    Given get answers of typeql read query
+      """
+      match
+       $x isa person;
+       $friend in friends-of($x);
+      """
+    Then answer size is: 25
+
+
+  # TODO: Do we want to keep this? Taken from value-predicate
```
flyingsilverfin, line 67: I feel like having more tests isn't bad as long as it's in the right place!

query/functions/basic.feature
```diff
@@ -0,0 +1,153 @@
...
+      """
+    Given transaction commits
+
+    Given connection open read transaction for database: typedb
+    Given get answers of typeql read query
+      """
+      match $x1, $x2 in message-successor-pairs();
+      """
+    # the (n-1)th triangle number, where n is the number of replies to the first post
+    Then answer size is: 10
```
flyingsilverfin, line 128: Anyway this reads as pretty awesome!

query/functions/basic.feature
```diff
@@ -0,0 +1,153 @@
...
+
+
+  Scenario: A function can return a value derived from an expression
```
flyingsilverfin, line 131: Ok so i think we're missing some base cases right?

1. return a stream of non-tuple instances (-> `{person}`)
2. return a stream of non-tuple values (-> `{long}`)
3. return a n-tuple of instances + values (probably just mix in 1 test should be ok) (`{ x, y, z}`)
4. return a singular instance and value `-> person | -> long`
5. return tuples of instances and values `-> long, long, long`
6. Future (could write them now?) Optional variants of the above... ` -> { long?, person }` `-> person?` `-> long?` -> `long?, long, long`

query/functions/usage.feature
```diff
@@ -0,0 +1,471 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#noinspection CucumberUndefinedStep
+Feature: Function call positions behaviour
```
flyingsilverfin, line 6: Great - though it's probably not just this. Perhaps we have two files? `definition.feature`, and then you can roll in stuff to do with calling them with invalid nr of arguments etc into `basic.feature`?

query/functions/structure.feature
```diff
@@ -0,0 +1,150 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#noinspection CucumberUndefinedStep
+Feature: Tests for various shapes of function bodies
```
flyingsilverfin, line 6: Don't need to write `Tests for...`! We don't have that anywhere else? Cue - filename. I'd call it `Feature: Function Body Structure`

query/functions/signature.feature
```diff
@@ -0,0 +1,219 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#noinspection CucumberUndefinedStep
+Feature: Effects a function signature have on the caller
```
flyingsilverfin, line 6: Function Call Validation? (and update file name)

query/functions/usage.feature
flyingsilverfin: Acutally the function call stuff can go into the one you've called `signature`, which is also includes mostly function call validation.

flyingsilverfin (changes requested):

query/functions/usage.feature
krishnangovindraj: Definition yes.
Basic is basic function execution.

query/functions/signature.feature
krishnangovindraj: I've got some where I check the body against the signature too.

flyingsilverfin (approved):

PR #323, opened 2024-12-24
Fix most of the query/language/insert tests

farost:
## Usage and product changes
We fix most of the `query/language/insert.feature` tests by updating TypeQL statements to 3.0 and removing outdated scenarios (e.g. attributes owning attributes). After these fixes, the result of running these tests is `121 scenarios (98 passed, 23 failed)`.

### DIFF
--- a/query/language/insert.feature
+++ b/query/language/insert.feature
@@ -19,7 +19,7 @@ Feature: TypeQL Insert Query
 
       entity person
         plays employment:employee,
-        owns name  @card(0..),
+        owns name @card(0..),
         owns age,
         owns ref @key,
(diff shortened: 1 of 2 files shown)


query/language/insert.feature
```diff
@@ -495,6 +511,7 @@ Parker";
       match $p has name "Spiderman";
       """
     Given answer size is: 0
+    # TODO: error "IsaStatementForInputVariable"
```
farost, line 514: Added some of the errors that I don't understand -> looking like bugs

flyingsilverfin (approved):

PR #327, opened 2025-01-28
Add tests for relates specialization definitions, redefinitions, and undefinitions

farost:
## Usage and product changes
Add tests for relates specialization definitions, redefinitions, and undefinitions.
Previously, the tests relied on Concept API tests too much and didn't check multiple cases which ended up broken in https://github.com/typedb/typedb/issues/7322 because of the implementation details. Now, we add more explicit cases for validations.

### DIFF
--- a/query/language/redefine.feature
+++ b/query/language/redefine.feature
@@ -572,7 +572,7 @@ Feature: TypeQL Redefine Query
     Then transaction commits
 
     When connection open schema transaction for database: typedb
-    Then typeql schema query; fails with a message containing: "'empty-relation:empty-role' should be inherited without specialization"
+    Then typeql schema query; fails with a message containing: "'empty-relation:empty-role' must be inherited without specialization"
       """
       redefine internship relates intern as empty-role;
       """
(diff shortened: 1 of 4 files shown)


query/language/redefine.feature
```diff
@@ -543,6 +543,41 @@ Feature: TypeQL Redefine Query
       """
 
 
+  Scenario: redefining specialisation to an inherited not specialising relates is allowed
```
flyingsilverfin, line 546: Title should use implicit in the second half? Or should we  leave that pretty generic with 'specialisation' as the term.

query/language/undefine.feature
```diff
@@ -651,6 +651,29 @@ Feature: TypeQL Undefine Query
       """
 
 
+  Scenario: undefining specialising relates as if it was a real relates is not allowed
```
flyingsilverfin, line 654: tbh I think `undefining relates as if it were a declared relates is not allowed`, or?

flyingsilverfin (approved):

query/language/redefine.feature
farost: Could be either specialising or implicit, I believe. I'll change it.

query/language/undefine.feature
farost: Implicit and explicit. I'll change it.

PR #330, opened 2025-02-03
Tests for commit time cardinality validation based on schema changes

farost:
## Usage and product changes
Add new concept migration tests for commit time cardinality validations for cardinality schema changes and change already existing tests to expect errors on commits instead of at operation time.

Additionally, it contains a case reproducing https://github.com/typedb/typedb/issues/7317.

## Implementation
All old tests failing on "<action>;fails" are rewritten to "<action> -> commit; fails with a message containing ... card ...". There was only a limited amount of places that checked other failures "just in case", they are left unchanged. Additionally, more specific error message checks were introduced in the migration tests.

New tests are written only for owns as the logic is written symmetrically. However, they are written in the way they can be reproduced for plays and relates if needed.

### DIFF
--- a/concept/migration/data-validation.feature
+++ b/concept/migration/data-validation.feature
@@ -4706,6 +4706,81 @@ Feature: Data validation
     Then transaction commits
 
 
+  Scenario: Cardinality is not validated if cardinality is unset to the default value with the owns unsetting
+    Given create attribute type: attr0
+    Given attribute(attr0) set value type: string
+    Given attribute(attr0) set annotation: @abstract
+    Given create attribute type: attr1
+    Given attribute(attr1) set supertype: attr0
+
+    Given create entity type: ent
+    Given entity(ent) set owns: attr0
+    Given entity(ent) set owns: attr1
+
+    Given entity(ent) get owns(attr0) set annotation: @card(0..)
+    Given entity(ent) get owns(attr1) set annotation: @card(2..)
+    Given transaction commits
+
+    Given connection open write transaction for database: typedb
+    Given $ent = entity(ent) create new instance
+    Given $attr0 = attribute(attr1) put instance with value: "attr0"
+    Given $attr1 = attribute(attr1) put instance with value: "attr1"
+    Given entity $ent set has: $attr0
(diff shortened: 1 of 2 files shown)


concept/migration/data-validation.feature
```diff
@@ -4157,3 +4583,486 @@ Feature: Data validation
     When $rel2 = relation(rel2) get instance with key(ref): rel2
     When relation $rel2 add player for role(role2): $ent2
     Then transaction commits
+
+
+  Scenario: Cardinality is not validated if instance is deleted
```
farost, line 4588: New tests start here. They contain a couple of simplified cases to ensure they are covered in the future without worrying. Additionally, there are tests covering combinations of `unset has` + `unset owns` + `attribute unset supertype` + `entity unset supertype`. 

I did cover it inside the long tests checking it at operation time, but moving the checks to commit time introduced new risks of possibly forgetting to mark specific super-super attribute types as affected when a related owns is unset or their subtype is changed, so I thought about these cases and wrote specific tricky tests to catch these failures. With these tests, I am more confident that at least 99% of schema modification cases is correctly validated.

concept/migration/data-validation.feature
```diff
@@ -4157,3 +4583,486 @@ Feature: Data validation
...
+      | 1..   | 0..   | attr1           | ; fails with a message containing: "@card(1..)"  |
+      | 0..   | 1..   | attr0           | ; fails with a message containing: "@card(1..)"  |
+      | 0..   | 1..   | attr1           |                                                  |
+      | 1..   | 1..1  | attr0           | ; fails with a message containing: "@card(1..1)" |
+      | 1..   | 1..1  | attr1           | ; fails with a message containing: "@card(1..)"  |
+      | 0..   | 0..   | attr0           |                                                  |
+      | 0..   | 0..   | attr1           |                                                  |
+
+
+  Scenario Outline: Cardinality is correctly revalidated when owns is specialised on a supertype with @card(<card>)
```
farost, line 5014: https://github.com/typedb/typedb/issues/7317

flyingsilverfin (approved):

PR #333, opened 2025-02-04
Update tests for attributes: concrete supertypes, independent attributes, deletes before and after commits

farost:
## Usage and product changes
Update tests for concrete attribute supertypes:
* Concept: remove expected failures if a concrete supertype is set for an attribute type.
* Extend cardinality tests to check instances of attributes and super attributes.
* Extend uniqueness checks for instances of the types with unique declared on types.
* Add match and match-insert tests to cover basic definitions through queries and cover insertions and retrievals.
* Add a match-delete test to cover deletions (that it's not somehow optimized to skip supertypes for whatever reason..).
* Separately test match-insert and match-delete with multiple object subtypes and different ownerships to check type inference (some already fixed bugs were found).
* Additional more complicated cases for independent and dependent attributes (for other found bugs).

## Implementation
Some of the tests are quite big due to permutations of actions. While it may seem redundant, these specific tests unveiled a couple of tricky bugs in the server, so I preserved it the way it's written for better coverage.


concept/migration/BUILD
```diff
@@ -7,7 +7,6 @@ package(default_visibility = ["//visibility:public"])
 load("@typedb_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
 
 exports_files([
-    "function.feature",
```
farost, line 10: Cleanup of unused stuff. We do need migration tests for functions, not sure if it's covered in current function tests. If it is (e.g., deletion of a type when it's referenced in a function), then let's leave it there or move it here. If not, let's decide where to put it.

I can return this file with a more explicit TODO with a github task for us to fill. Or I can even write these tests lol

farost: @krishnangovindraj

concept/thing/roleplayer.feature
```diff
@@ -206,3 +206,95 @@ Feature: Concept Role Players
     When $rel0 = relation(rel0) create new instance with key(ref): rel0
     Then relation $rel0 set players for role(rol0[]): [$ent1, $ent2]; fails
 
+  Scenario: Links can be unset before and after commits
```
farost, line 209: I added this test to replicate the bug-related test from `has`. However, it works differently for roleplayers. I decided to leave it here to pin this behavior (see the comment below)

concept/thing/roleplayer.feature
```diff
@@ -206,3 +206,95 @@ Feature: Concept Role Players
...
+    When $p = entity(person) get instance with key(name): "p"
+    When $e = relation(employment) get instance with key(date): 2025-02-12
+    Then relation $e exists
+    Then relation $e get players for role(employee) contain: $p
+    Then entity $p get relations contain: $e
+    When relation $e add player for role(employee): $p
+    Then relation $e get players for role(employee) contain: $p
+    Then entity $p get relations contain: $e
+    When relation $e remove player for role(employee): $p
+    # 3 are inserted, 1 is removed -> 2 is left
```
farost, line 284: While `unset has` unsets has completely, this removes a single player, so we need to call it multiple times.

concept/migration/BUILD
krishnangovindraj: `query/function/definition.feature` is where I have one such test. (Scenario: If a modification of the schema causes a stored function to become invalid, the modification is blocked at commit time)
Can you move the commented out scenarios there?

farost: Okay! Thanks

concept/thing/relation.feature
```diff
@@ -205,7 +205,89 @@ Feature: Concept Relation
...
+    Then relation $m does not exist
+    Then relation(marriage) get instances is empty
+    When $m = relation(marriage) create new instance with key(license): m
+    When $a = entity(person) create new instance with key(username): alice
+    When relation $m add player for role(wife): $a
+    When delete entity: $a
+    Then entity $a is deleted: true
+    Then relation $m add player for role(wife): $a; fails
+
+  Scenario: Relaton cannot have multiple layers of set role players
```
dmitrii-ubskii, line 244:
```suggestion
  Scenario: Relation cannot have multiple layers of set role players
```

concept/type/attributetype.feature
```diff
@@ -465,7 +465,7 @@ Feature: Concept Attribute Type
 # @abstract
 ########################
 
-  Scenario: Attribute types can be set to abstract
+  Scenario: @abstract annotation can be set and unset for attribute types
```
dmitrii-ubskii, line 468: I think we're missing a test for trying to unset `@abstract` on an attribute type that has no set `value`. Unless it's in a separate scenario, of course.

query/language/delete.feature
```diff
@@ -1011,8 +1101,8 @@ Feature: TypeQL Delete Query
       friendship owns timespan;
       """
     Given transaction commits
-
     Given connection open write transaction for database: typedb
+
     Given typeql write query
```
dmitrii-ubskii, line 1106: The spacing made more sense before

concept/type/attributetype.feature
farost: It's `Attribute types cannot unset @abstract annotation without value type`

dmitrii-ubskii (approved):
Note for future reference: a lot of scenarios are now _really_ long and hard to read. Breaking them into smaller scenarios would be preferable.

query/language/delete.feature
farost: Agree, I think I inserted something and then decided to remove it.

PR #334, opened 2025-02-12
Add test for duplicate inserts & redundant deletes

krishnangovindraj:
## Usage and product changes
Add test for duplicate inserts & redundant deletes

## Implementation
Add test for duplicate inserts & redundant deletes

### DIFF
--- a/query/language/delete.feature
+++ b/query/language/delete.feature
@@ -505,22 +505,11 @@ Feature: TypeQL Delete Query
     """
     Then transaction commits
 
-    # We're not particular about this:
-    Given connection open write transaction for database: typedb
-    When typeql write query; fails with a message containing: "Write execution failed due to a concept write error"
-    """
-    match
-      $e isa employment, links (employee: $p);
-    delete
-      links (employee: $p) of $e;
-    """
-    Given transaction closes
-
     Given connection open write transaction for database: typedb
     When typeql write query
     """
     match
-      $e isa! employment, links (employee: $p);
+      $e isa employment, links (employee: $p);
     delete
       links (employee: $p) of $e;
     """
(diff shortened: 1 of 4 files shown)


query/language/delete.feature
```diff
@@ -792,6 +792,51 @@ Feature: TypeQL Delete Query
...
+       """
+    Given transaction commits
+
+    Given connection open write transaction for database: typedb
+    When get answers of typeql read query
+    """
+    match
+      $e isa internship, links (intern: $p);
+    """
+    Then answer size is: 1
```
farost, line 821: It could be read easier if we actually checked what are the answers. For example, their labels. At least it's easier for me when I skim old (even my) tests

query/language/insert.feature
```diff
@@ -208,6 +208,17 @@ Feature: TypeQL Insert Query
       insert $x isa factory;
       """
 
+
+  Scenario: A single variable may not have multiple isa constraints in an insert stage
+    When typeql write query; fails with a message containing: "Found multiple insert statements for the variable"
```
farost, line 213: Not sure if `multiple insert statements` is enough for a user to identify that the problem is in `isa`s! But it might be reviewed in the server's PR, seeing the whole stack trace. It should probably highlight the code region causing the error

farost (approved):

query/language/delete.feature
```diff
@@ -1430,6 +1466,8 @@ Feature: TypeQL Delete Query
     Then answer size is: 0
 
 
+  # TODO: 3.x: Needs finer write type-checks
+  @ignore
```
krishnangovindraj, line 1470: This is quite bad.

PR #338, opened 2025-02-17
Introduce tests for update queries

farost:
## Usage and product changes
Introduce tests for update queries. These queries are used to make sure that a type capability has a specific value: either an attribute or a role player.
 
The old tests in the `update.feature` using `delete + insert` were either moved into the `pipelines` or updated to contain two compared versions: `delete + insert` and `update`. In the meantime, the test file is now full of new tests for the new feature:
1. Validations of incorrect usages of the queries
2. Updating `has`: introduction of a `has` without having an old capability, updates of existing capabilities, validation for a multi-layered attribute `sub` hierarchy, cardinality validations, and more. 
3.  Updating `links`: same as in the previous point, but with specifics of roleplayers.
4. Combinations and edge cases.

### DIFF
--- a/query/language/delete.feature
+++ b/query/language/delete.feature
@@ -2695,3 +2695,31 @@ Feature: TypeQL Delete Query
       delete
         $x;
       """
+
+
+  Scenario: deleting an anonymous variable errors
+    Then typeql write query; fails with a message containing: "anonymous"
+      """
+      delete
+        $_ of $_;
+      """
+    When transaction closes
+
+    When connection open write transaction for database: typedb
+    Then typeql write query; fails with a message containing: "anonymous"
+      """
+      insert
+        $p isa person, has name "John";
+      delete
+        $_ of $p;
+      """
+    When transaction closes
+
+    When connection open write transaction for database: typedb
+    Then typeql write query; fails with a message containing: "anonymous"
+      """
+      insert
+        $n isa name "John";
+      delete
+        $n of $_;
+      """
--- a/query/language/delete.feature
+++ b/query/language/delete.feature
@@ -1101,8 +1101,8 @@ Feature: TypeQL Delete Query
       friendship owns timespan;
       """
     Given transaction commits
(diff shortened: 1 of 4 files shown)


query/language/delete.feature
```diff
@@ -2695,3 +2695,31 @@ Feature: TypeQL Delete Query
       delete
         $x;
       """
+
+
+  Scenario: deleting an anonymous variable errors
```
farost, line 2700: It failed. I added a fix. I test the fix.

query/language/update.feature
```diff
@@ -17,140 +17,2906 @@ Feature: TypeQL Update Query
...
+#      """
+#      match
+#        $p0 isa person, has ref 0, has balance $b0;
+#        $p1 isa person, has ref 1, has balance $b1;
+#      update
+#        $p1 has balance get_balance($p0);
+#      """
+
+
+  Scenario Outline: Cannot update has with cardinality higher than 1: @card(<card>)
```
flyingsilverfin, line 736: good test!

query/language/update.feature
```diff
@@ -17,140 +17,2906 @@ Feature: TypeQL Update Query
...
+      """
+      match
+        $p isa person;
+      update
+        $p has first-name "Elon";
+        $p has ref 0;
+      """
+
+
+  Scenario: Update queries check the actual has cardinality
```
flyingsilverfin, line 1394: Isn't this subsumed in` Cannot update has with cardinality higher than 1: @card(<card>)`?

flyingsilverfin (approved):

flyingsilverfin:
Left 1 comment to answer

query/language/update.feature
farost: It can be extended so it becomes a subset. But I check two separate things:
Test 1. Different cards are rejected the same way
Test 2. If card is changed, we read this change without commits

PR #339, opened 2025-02-21
Update update bdd for static type checks

krishnangovindraj:
## Usage and product changes
Update update bdd for static type checks

## Implementation
Update update bdd for static type checks

### DIFF
--- a/query/language/update.feature
+++ b/query/language/update.feature
@@ -1919,16 +1919,6 @@ Feature: TypeQL Update Query
       | key:ref:0 | key:ref:1 | label:aged-friendship:younger-friend |
       | key:ref:0 | key:ref:1 | label:aged-friendship:older-friend   |
 
-    # This would not actually be inserted as it's specialised, but we should also see this error first
-    Then typeql write query; fails with a message containing: "cardinality should not exceed 1"
-      """
-      match
-        $f isa aged-friendship, has ref 0;
-        $p isa person, has ref 0;
-      update
-        $f links (friend: $p);
-      """
-
 
   Scenario Outline: Links for card(<card>) can be updated
     Given transaction closes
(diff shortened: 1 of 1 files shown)


query/language/update.feature
```diff
@@ -1919,16 +1919,6 @@ Feature: TypeQL Update Query
       | key:ref:0 | key:ref:1 | label:aged-friendship:younger-friend |
       | key:ref:0 | key:ref:1 | label:aged-friendship:older-friend   |
 
-    # This would not actually be inserted as it's specialised, but we should also see this error first
```
farost, line 1922: Why do you just remove a case? Update the error message.

farost (changes requested):

query/language/update.feature
krishnangovindraj: In person discussion, adapted test.

query/language/update.feature
```diff
@@ -1918,12 +1918,11 @@ Feature: TypeQL Update Query
...
       | key:ref:0 | key:ref:1 | label:aged-friendship:younger-friend |
       | key:ref:0 | key:ref:1 | label:aged-friendship:older-friend   |
-
-    # This would not actually be inserted as it's specialised, but we should also see this error first
+    # We don't need instances since it's a compile-time check
     Then typeql write query; fails with a message containing: "cardinality should not exceed 1"
       """
       match
-        $f isa aged-friendship, has ref 0;
+        $f isa! friendship, has ref 0;
```
farost, line 1925: We are still interested in the original case I tested. What will happen if a user decides to write this "general" update, expecting it to come up with a correct role type based on the `$f isa`? This should be at least some error.
I'd expect myself to have other tests written with correct cardinalities, but I don't see an issue in having such a small case here, too.

krishnangovindraj: > expecting it to come up with a correct role type based on the `$f isa`
We don't support this currently. We error.

Also, I think it's covered under: `Update queries are validated similarly to insert queries and do not allow creating abstract links instances`

farost: Well, the last point is enough.

farost (approved):

PR #344, opened 2025-02-26
Adjust tests to allow abstract relation types having zero role types

farost:
## Usage and product changes
Adjust tests to allow abstract relation types have zero role types. Update some of the tests to expect validation errors at operation time instead of commits.
Additionally, add more cases for role type modification to catch bugs related to incorrect specialized role types' deletions.

### DIFF
--- a/concept/type/relationtype.feature
+++ b/concept/type/relationtype.feature
@@ -94,7 +94,12 @@ Feature: Concept Relation Type and Role Type
     Then get role types do not contain:
       | parentship:parent |
       | parentship:child  |
-    Then relation(marriage) delete role: spouse; fails
+    When transaction commits
+    When connection open schema transaction for database: typedb
+    When relation(marriage) delete role: spouse
+    Then relation(marriage) get relates is empty
+    Then transaction commits; fails
+    When connection open schema transaction for database: typedb
     When relation(marriage) create role: husband
     When relation(marriage) create role: wife
     When relation(marriage) delete role: spouse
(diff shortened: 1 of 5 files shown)


query/language/undefine.feature
```diff
@@ -2316,15 +2316,144 @@ Feature: TypeQL Undefine Query
...
+    When transaction closes
+
+    When connection open schema transaction for database: typedb
+    Then typeql schema query
+      """
+      undefine
+      as parent from fathership relates father;
+      @card from fathership relates father;
+      relates father from fathership;
+      # @abstract from fathership; # fathership does not have roles at this point. we may want to allow it or error...
```
farost, line 2453: It can be a little strange that we can't undefine the abstract annotation here. I don't mind too much, but highlighted it here in case we will want to change it. I can already remove the check if you prefer to. 

Although it's funny that I've found the additional bug (which is covered in concept bdds) specifically because I started thinking about these permutations and added more separate cases above! I hate this fact that I just spam tests sometimes, but they do actually work well......

concept/type/relationtype.feature
```diff
@@ -94,11 +94,17 @@ Feature: Concept Relation Type and Role Type
     Then get role types do not contain:
```
farost, line 94: Old tests: mostly just changed commit/operation time validations and sometimes added trickier cases.

concept/type/relationtype.feature
```diff
@@ -789,6 +792,273 @@ Feature: Concept Relation Type and Role Type
     Then relation(family) get relates(parentship:parent) is implicit: false
     Then relation(family) get constraints for related role(parentship:parent) do not contain: @abstract
 
+  Scenario: Relation type specialisation is preserved if the role type is specialised at least once after specialisation unsets
```
farost, line 795: These tests cover that implicit relates are correctly cleaned up after... actions. And do not actually depend on other relation types (there was a related bug in the server).

concept/type/relationtype.feature
```diff
@@ -1513,6 +1800,74 @@ Feature: Concept Relation Type and Role Type
     When connection open schema transaction for database: typedb
     Then relation(rel00) delete role: role00; fails
 
+  Scenario: Deleting a relation type also deletes its declared role types
```
farost, line 1803: This scenario has been a great help, surprisingly

flyingsilverfin (approved):

PR #346, opened 2025-03-03
Add behaviour tests for 'put'

krishnangovindraj:
## Usage and product changes
Add behaviour tests for 'put'

## Implementation
Add behaviour tests for 'put'

### DIFF
--- a/query/language/put.feature
+++ b/query/language/put.feature
@@ -19,7 +19,6 @@ Feature: TypeQL Insert Query
         plays employment:employee,
         owns name @card(0..),
         owns age,
-        owns ref @key,
         owns email @unique @card(0..);
 
       entity company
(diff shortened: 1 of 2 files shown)


query/language/put.feature
```diff
@@ -0,0 +1,505 @@
...
+
+  ####################
+  #  Put Instances  #
+  ###################
+
+  Scenario: Putting an entity will create it if no entity exists, or return all matching entities
+    Given connection open write transaction for database: typedb
+    # Verify put does insert if not exists
+    When get answers of typeql read query
+    """
```
farost, line 53: A little hard to read when the tabs are different for match/write queries

query/language/put.feature
```diff
@@ -0,0 +1,505 @@
...
+      """
+    Then answer size is: 1
+    When get answers of typeql read query
+    """
+    match $x isa person;
+    """
+    Then answer size is: 1
+
+    # Check put returns all matching answers by inserting one more person
+    Given get answers of typeql write query
```
farost, line 95: Could insert the same check after this step as well

farost (changes requested):
I'd like to see more

query/language/put.feature
```diff
@@ -0,0 +1,588 @@
...
+    """
+    Then answer size is: 1
+    When get answers of typeql read query
+    """
+    match $p isa person, has age $age;
+    """
+    Then answer size is: 2
+
+
+  Scenario: A match-put can be used to create an entity to own an existing attribute
```
krishnangovindraj, line 261: Test added on feedback.

query/language/put.feature
```diff
@@ -0,0 +1,588 @@
...
+    """ match $p isa person, has name "bob"; """
+    Then answer size is: 1
+    When get answers of typeql write query
+    """ put $p isa child, has name "bob"; """
+    Then answer size is: 1
+    When get answers of typeql read query
+    """ match $p isa person, has name "bob"; """
+    Then answer size is: 2
+
+  Scenario: Putting an ownership when a subtype is already owned does nothing.
```
krishnangovindraj, line 407: Test added on feedback

krishnangovindraj (commented):
replies

query/language/put.feature
```diff
@@ -0,0 +1,505 @@
...
+    When get answers of typeql read query
+    """
+      match $p isa person, has residential-address "9, Downing Street";
+    """
+    Then answer size is: 1
+    When typeql write query; fails with a message containing: "Type-inference derived an empty-set for some variable"
+    """
+      put $p isa person, has address "9, Downing Street";
+    """
+
```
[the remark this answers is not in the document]
krishnangovindraj: Good idea: 
https://github.com/typedb/typedb-behaviour/pull/346#discussion_r1978947243

query/language/put.feature
```diff
@@ -0,0 +1,505 @@
...
+    match $p isa person, has name "bob", has email $email;
+    """
+    Then uniquely identify answer concepts
+      | email                     |
+      | attr:email:bob@email.com  |
+      | attr:email:bob@typedb.com |
+
+
+  Scenario: Putting an ownership when a subtype already owns the attribute does nothing.
+    # TODO: Add test for owning subtypes when we allow non-abstract super-attributes?
```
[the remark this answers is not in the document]
krishnangovindraj: ah ok: https://github.com/typedb/typedb-behaviour/pull/346#discussion_r1978952028

query/language/put.feature
```diff
@@ -0,0 +1,505 @@
...
+    Then answer size is: 1
+    Then uniquely identify answer concepts
+      | cname            | pname           |
+      | attr:name:typedb | attr:name:alice |
+
+
+  ####################
+  #  Validation      #
+  ####################
+  Scenario: Concepts in a put stage must either be an input or be insertable in the put stage.
```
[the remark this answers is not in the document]
farost: Plus, sometimes it's not limited by the syntax, and we need to cover it additionally in the IR. You developed the feature and you are not sure if it's a legal syntax - that's the perfect sign that we need these tests.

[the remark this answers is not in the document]
farost: I'm not talking about covering every permutation, but covering each group of operations (expressions, types definitions, etc) is an up 10 minutes job.

query/language/put.feature
farost: It just lets you read the test easier, from my point of view. You go like:
* Alright, we insert... 
* Ok, we get 2 persons after the insert...
* Right, we put... and...
* ... and there is still 2 persons 

It's not a complicated thing since you know what to expect from inserts, but it makes reading large tests files / debugging possible errors easier.

query/language/put.feature
```diff
@@ -0,0 +1,505 @@
...
+    When get answers of typeql write query
+    """
+    put $x isa person, has ref 0;
+    """
+    Then uniquely identify answer concepts
+      | x         |
+      | key:ref:0 |
+    When get answers of typeql read query
+    """
+    match $x isa! $t; $t sub person;
```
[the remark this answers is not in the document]
farost: I just didn't find this query easy to read to verify that this test covers it completely. I don't mind leaving it as is, just mentioned that adding/changing this to counts might've simplified the readability of the test (what i'd do, not what I expect from you)

query/language/put.feature
```diff
@@ -0,0 +1,588 @@
...
+
+  Scenario: If the entire pattern does not match, put will insert the whole pattern. To insert a partial pattern, pipeline puts.
+    Given connection open write transaction for database: typedb
+    When typeql write query
+    """
+    insert $p1 isa person, has name "alice", has email "alice@email.com";
+    insert $p2 isa person, has name "bob", has email "bob@email.com";
+    """
+    Given transaction commits
+
```
farost, line 312: Can remove one of the newlines

query/language/put.feature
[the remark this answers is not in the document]
farost: let's remove the comment, then

query/language/put.feature
farost: Could also verify that a subtype is inserted if you put it when a supertype is already owned

farost (commented):
Comments

query/language/put.feature
krishnangovindraj: We do that at the end actually.

query/language/put.feature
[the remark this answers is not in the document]
krishnangovindraj: Feels like noise in a BDD setup honestly. But I'm unsure where else to put it.
I agree we lack test coverage so it needs to be covered somewhere.

query/language/put.feature
krishnangovindraj: > Ok, we get 2 persons after the insert...
Misleading. We don't "get two persons after the insert". We inserted 2 persons. There may now be n >= 2 persons in the database.

farost: That's the point!!!!! We didn't insert 2 persons, we inserted 1, and we get 2 total if we run a read query after this Given! Otherwise, we need to additionally think: "Hmmm what do we get after this insert?", while the file is about puts. You probably understood my message from a different angle, but it still works
I'm not going to insist, I just find it much easier to follow

farost (approved):

query/language/put.feature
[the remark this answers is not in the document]
flyingsilverfin: I agree it's good to have a test to verify things like expressions would not work in put, since it's also an insert in disguise :)

flyingsilverfin (approved):

PR #352, opened 2025-03-27
Ensure function arguments are passed according to index

krishnangovindraj:
## Usage and product changes
Add regression test (typedb/typedb#7407) to ensure function arguments are passed according to index

## Implementation
Add regression test (typedb/typedb#7407) to ensure function arguments are passed according to index


krishnangovindraj:
Close in favour of : https://github.com/typedb/typedb-behaviour/pull/354

PR #353, opened 2025-03-28
Add tests to ensure variables input to a stage cannot be assigned to

krishnangovindraj:
## Usage and product changes
Add tests to ensure variables input to a stage cannot be assigned to

## Implementation
* Updates a test in expressions, adds one in functions usage.


krishnangovindraj:
Close in favour of : https://github.com/typedb/typedb-behaviour/pull/354

PR #354, opened 2025-03-28
Add regression tests for various fixes around variable usage

krishnangovindraj:
## Usage and product changes
Add regression tests for various fixes in https://github.com/typedb/typedb/pull/7410:
* Extend the check for empty type-annotations to include label vertices
* Store function call arguments in a Vec, so they are ordered by argument index. 
* Flag assignments to variables which were already bound in previous stages
* Flag anonymous variables in unexpected pipeline stages instead of assuming named & unwrapping.


query/functions/usage.feature
```diff
@@ -163,3 +164,48 @@ Feature: Function Usage
       | key:ref:0 |
     Given transaction closes
 
+
+  Scenario: Function arguments are passed properly according to index (typedb#7407)
```
farost, line 168: It feels like a common test to have, I don't think that the issue/pr reference is required here.

farost (approved):

query/functions/usage.feature
krishnangovindraj: Yeah but I may not cover every case.

PR #355, opened 2025-03-28
Test nested negation inputs

dmitrii-ubskii:
## Usage and product changes

Reproduces the crash fixed by https://github.com/typedb/typedb/pull/7411

### DIFF
--- a/query/language/negation.feature
+++ b/query/language/negation.feature
@@ -605,5 +605,28 @@ Feature: Negation Resolution
     Then answer size is: 3
 
 
+  Scenario: Negation inputs are handled correctly
+    Given connection open schema transaction for database: typedb
+    Given typeql write query
+      """
+      insert
+      $x isa person;
+      $c isa company;
+      $e isa employment, links (employee: $x, employer: $c);
+      """
+    Given transaction commits
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        not {
+          $x isa person;
+          not {
+            (employee: $x);
+          };
+        };
+      """
+
+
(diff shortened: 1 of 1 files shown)


query/language/negation.feature
```diff
@@ -605,5 +605,31 @@ Feature: Negation Resolution
...
+      match
+        not { # Must be in a negation so that `$x` is not selected for the entire match stage.
+          $x isa person;
+          not {
+            (employee: $x);
+            # The planner prefers ordering `$_ role-name employee` (which is free) before `$_ links $_:$x` (which has to hit disk).
+            # If `$x` is not marked as input to the negation during lowering, it is not selected by the role name step, and therefore removed from the row.
+            # The links executor then crashes as it expects `$x` to be bound.
+          };
+        };
```
dmitrii-ubskii, line 630: I'm not sure about these explanatory comments, but I wanted to explain that everything in this test query is there for a specific reason, for future reference.

query/language/negation.feature
```diff
@@ -605,5 +605,31 @@ Feature: Negation Resolution
...
+      """
```
krishnangovindraj, line 631: Do you want to verify the answer size in a `then` just to make sure the executor actually does something?

krishnangovindraj (approved):
one comment. Looks good.

query/language/negation.feature
dmitrii-ubskii: It's a negation, it can only ever output one zero-width row.

krishnangovindraj: Makes sense

krishnangovindraj (approved):

PR #361, opened 2025-04-17
Add http-specific behaviour scenarios

farost:
## Usage and product changes
Add scenarios for methods exclusive for the TypeDB HTTP Endpoint:
* Query options
* Non-authorized and wrongly authorized calls
* One-shot queries and their transaction/query options

### DIFF
--- a/driver/http/BUILD
+++ b/driver/http/BUILD
@@ -1,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package(default_visibility = ["//visibility:public"])
+
+load("@typedb_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+
+files = [
+    "http.feature",
+]
+
+filegroup(
+    name = "features",
+    srcs = files,
+)
+
+exports_files(files)
+
+checkstyle_test(
+    name = "checkstyle",
+    include = glob(["*"]),
+    license_type = "mpl-header",
+)
--- a/driver/http/http.feature
+++ b/driver/http/http.feature
@@ -28,6 +28,14 @@ Feature: TypeDB HTTP Endpoint
   # AUTHENTICATION #
   ##################
 
+  Scenario: Authentication is invalidated after the authentication token TTL
+    Then connection create database: typedb2
+    When wait authentication token expiration
+    Then connection create database: typedb3; fails with a message containing: "Invalid token"
+
+    When connection opens with default authentication
+    Then connection create database: typedb3
+
   Scenario: Database methods are not available without authentication
     When connection closes
     Then connection is open: false
--- a/driver/http/http.feature
+++ b/driver/http/http.feature
@@ -139,6 +139,174 @@ Feature: TypeDB HTTP Endpoint
       match entity $x;
       """
 
+  #################################
+  # CONCEPT ROW ANSWERS STRUCTURE #
+  #################################
+
+  Scenario: Transaction read row queries return all concepts correctly
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+      """
+      define
+        entity person, plays parentship:parent, owns name;
+        attribute name value string;
+        relation parentship, relates parent;
+      """
+    Given typeql write query
+      """
+      insert
+        $p isa person, has name 'John';
+        $pp isa parentship, links ($p);
+      """
+    Given transaction commits
+
+    When connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        $entity isa $entity-type, has $attribute-type $attribute;
+        $relation isa $relation-type, links ($entity);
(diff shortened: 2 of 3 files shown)


driver/http/BUILD
```diff
@@ -0,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package(default_visibility = ["//visibility:public"])
+
+load("@typedb_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+
+files = [
+    "http.feature",
```
flyingsilverfin, line 10: Good idea, we can definitely also test http in different languages and from the server too!

driver/http/http.feature
```diff
@@ -0,0 +1,1140 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#noinspection CucumberUndefinedStep
+Feature: TypeDB HTTP Endpoint
+
+  Background: Open connection, create database
+    Given typedb starts
+    Given connection is open: false
```
flyingsilverfin, line 10: Does this make any sense for HTTP? Or are you trying to re-use steps

driver/http/http.feature
```diff
@@ -0,0 +1,1140 @@
...
+    Then answer column names are:
+      | p |
+      | a |
+    Then answer get row(0) get concepts size is: 2
+    Then answer get row(0) get entity(p) get type get label: person
+    Then answer get row(0) get entity by index of variable(p) get type get label: person
+    Then answer get row(0) get attribute(a) get type get label: name
+    Then answer get row(0) get attribute(a) get value is: "John"
+    Then answer get row(0) get attribute by index of variable(a) get type get label: name
+    Then answer get row(0) get attribute by index of variable(a) get value is: "John"
```
flyingsilverfin, line 312: since HTTP responses are JSON, why don't we use the string-based verification here like we do for Fetch?

driver/http/http.feature
farost: I was trying to re-use the steps. But it probably does not, I can remove this!

driver/http/http.feature
farost: I don't want to implement additional steps and implementations for them, to be honest... It's just a copypaste of the driver steps, but with one-shot queries.

farost: Okay, I was thinking about it a little. Yeah, it makes sense to **additionally* test their document structure. I won't require any additional steps implementations. I'll add a test for all the concept kinds.

driver/http/http.feature
flyingsilverfin: In your server-side implementation, you've got a different set of http_steps right? So there's no need to keep them exactly in-sync IMO. We can keep the http one simpler

driver/http/http.feature
flyingsilverfin: Well if you're going to do that too we don't really have to test these other access methods with so many steps right? Less code to maintain is always a valid goal

driver/http/http.feature
farost: I want to test the whole `driver.feature` file, which is more specific. And I need to verify that all the permutations of value types and values are parsed correctly. No need to substitute it. I just want to verify the structure of the output HTTP document, that's it.

driver/http/http.feature
farost: https://github.com/typedb/typedb-behaviour/pull/361#discussion_r2050719295

driver/http/http.feature
flyingsilverfin: ah right ok

farost: Added

driver/http/http.feature
```diff
@@ -0,0 +1,1307 @@
...
+      match
+        $entity isa $entity-type, has $attribute-type $attribute;
+        $relation isa $relation-type, links ($entity);
+        $relation-type relates $role-type;
+        let $value = $attribute;
+      """
+    Then answer size is: 1
+    Then answer type is: concept rows
+    Then answer type is not: concept documents
+    Then answer contains document:
```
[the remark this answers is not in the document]
farost: I'd avoid adding additional steps, to be honest. We have the checks for `concept rows` / `concept documents` above. This step can be read as `contains <JSON> document`, signaling that you expect a set of documents in the answer, and one of these documents is *this*. If the step is already declared, I don't see a need to duplicate it.

But, if you really don't like it, I don't mind adding another regex declaration to the server.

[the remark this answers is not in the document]
flyingsilverfin: Are you storing the answers in the same server-side field?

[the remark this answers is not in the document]
farost: Yes, but it should not matter here. I just want to make sure that, as a user, I see this document in the reply. That's how I see this.

flyingsilverfin (approved):

PR #362, opened 2025-04-24
Enable transaction and query options tests for drivers

farost:
## Usage and product changes
Move transaction-based scenarios for transaction options and the `include_instance_types` query option to the `driver.feature` file. Enable transaction options scenarios for all drivers.


driver/driver.feature
```diff
@@ -335,7 +334,7 @@ Feature: TypeDB Driver
       """
     When wait 4 seconds
     Then transaction is open: false
-    Then typeql read query; fails with a message containing: "no open transaction"
```
farost, line 338: The messages are different in two endpoints, and I don't want to merge them. It's hard to write this tests wrongly, and there is an additional check above, so it's fine to ignore the error mesage.

flyingsilverfin (approved):

flyingsilverfin (approved):

PR #363, opened 2025-04-28
Test concept deletions don't affect constraint deletions in the same stage

krishnangovindraj:
## Usage and product changes
Add test to check concept deletions don't affect constraint deletions in same stage

### DIFF
--- a/query/language/delete.feature
+++ b/query/language/delete.feature
@@ -2779,3 +2779,24 @@ Feature: TypeQL Delete Query
       delete
         $n of $_;
       """
+
+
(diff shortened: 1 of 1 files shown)


query/language/delete.feature
```diff
@@ -2779,3 +2779,24 @@ Feature: TypeQL Delete Query
       delete
         $n of $_;
       """
+
+
+  Scenario: Concept deletions do not cause trouble for constraint deletions in the same stage referencing that concept
```
farost, line 2784: It's a smart test!

farost (approved):

PR #366, opened 2025-05-15
Add tests for variable visibility

krishnangovindraj:
## Usage and product changes
Add tests for variable visibility to ensure subsequent stages do not see visibles which are disjunction local, deleted, or (3) reduced away.

## Implementation
Introduces tests in a newly introduced `validation.feature`

### DIFF
--- a/query/language/fetch.feature
+++ b/query/language/fetch.feature
@@ -2997,3 +2997,29 @@ Feature: TypeQL Fetch Query
         "also known as": "Jon"
       }
       """
+
+  ###############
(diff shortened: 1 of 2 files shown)


farost:
If the fix is initiated by a specific bug https://github.com/typedb/typedb/issues/7462 (you should also reference it in https://github.com/typedb/typedb/pull/7463), I'd like to see that we at least cover one of the cases in fetch. It will be easy to add to the `fetch.feature`. If not, then when do we plan to follow up with the fix?

PR #379, opened 2025-09-16
Update wrong host error message

krishnangovindraj:
The original message is in the source of the error but I can't find a way to get it out.

## Usage and product changes
Update wrong host error message to reflect the message returned by the updated dependency.


krishnangovindraj:
We decided to collect all the sources `.to_string()` instead.

PR #380, opened 2025-09-16
Update BDD for empty define

krishnangovindraj:
## Usage and product changes
Update connection & migration BDD to reflect empty `define` statements being valid (and used) TypeQL (typedb/typeql#412)

### DIFF
--- a/driver/connection.feature
+++ b/driver/connection.feature
@@ -130,6 +130,7 @@ Feature: Driver Connection
     Given connection has database: typedb
     Then connection get database(typedb) has schema:
     """
+    define
     """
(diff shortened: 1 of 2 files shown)


driver/connection.feature
```diff
@@ -130,6 +130,7 @@ Feature: Driver Connection
     Given connection has database: typedb
     Then connection get database(typedb) has schema:
     """
+    define
```
farost, line 133: Since we got a specific request regarding `define ;` (I assume the enclosing `;` is a possible part of the template), I would also cover `define ;` somewhere. Just to be completely sure.

farost (approved):

PR #392, opened 2025-11-19
Add behaviour scenarios for clustered TypeDB

farost:
## Usage and product changes

Add BDD tests for cluster driver support.

- **Cluster driver tests** (`driver/cluster.feature`): server discovery, roles, terms, single-server connection, server routing modes, failover retries configuration. Expected to run against a 3-node cluster.
- **Driver option tests** (`driver/connection.feature`): `request_timeout_millis` configuration and timeout failure.
- **Error message update** (`user.feature`): "User does not exist" → "User not found" to match the error message pattern.

## Implementation

New BDD steps requiring driver implementations:

- `connection has {int} server(s)` — query server list, assert count
- `connection primary server exists` — find server with primary role
- `connection get server({address}) exists / does not exist` — lookup by address
- `connection get server({address}) has term` — assert term is present
- `connection servers have roles:` — data table of expected roles
- `connection opens to single server with default authentication` — connect via single address
- `set operation server routing to: {routing}` — configure `auto` or `direct(address)` routing
- `set driver option {option} to: {value}` — configure `primary_failover_retries`, `request_timeout_millis`


driver/cluster.feature
```diff
@@ -0,0 +1,76 @@
...
+    Given connection is open: true
+
+  ###########
+  # SERVER #
+  ###########
+
+  Scenario: Driver can discover servers in a cluster
+    Then connection has 3 servers
+    Then connection primary server exists
+    Then connection get server(127.0.0.1:11729) exists
```
farost, line 23: Was thinking about something abstract like "Nth address", but it only added headache
It will be better when we dissolve `typedb-behaviour` and move the scenarios into `typedb-driver`, where these constants exists across the whole repo

flyingsilverfin: yes - this is ok though!

driver/cluster.feature
```diff
@@ -0,0 +1,76 @@
...
+    Then connection get server(127.0.0.1:21729) exists
+    Then connection get server(127.0.0.1:31729) exists
+
```
flyingsilverfin, line 26: just need a single newline between scenarios i think! Check what others are doing

driver/cluster.feature
```diff
@@ -0,0 +1,76 @@
...
+    Then connection has 3 servers
+    Then connection get server(127.0.0.1:11729) has term
+    Then connection get server(127.0.0.1:21729) has term
+    Then connection get server(127.0.0.1:31729) has term
+
+
+  Scenario: Driver can inspect server roles
+    Then connection has 3 servers
+    Then connection servers have roles:
+      | primary   |
```
flyingsilverfin, line 38: are these ordered? Do you want to key them on port or something, or is this just a set of roles that need to satisfied one each

flyingsilverfin (approved):
Approved w comments

driver/cluster.feature
farost: Nah. I enforced 2 newlines between the scenarios everywhere so we could logically split parts of each scenario with single newlines with the 3.x release, most of the files follow the same convention

driver/cluster.feature
farost: No, it's just a set. The server roles are not fixed, we can only enforce it by timely restarting them, but it's too much control for these BDDs. The keys won't work here, so we just basically ensure the counts

PR #400, opened 2026-01-27
Test datetime and string operators

dmitrii-ubskii:
## Usage and product changes

Add tests for new operators implemented in https://github.com/typedb/typedb/pull/7681


flyingsilverfin (approved):
Any failure cases to test?

PR #403, opened 2026-02-27
Migrate to Bazel 8

lolski:
## Motivation

Update Bazel version from 6.2 to 8.5.1.

The upgrade is done in a backwards-compatible way, such that "upstream" repositories that are yet to be upgraded may depend on this repository. This is done by preserving WORKSPACE and the deps.bzl loader files alongside the new Bazel 8 ones. Once every repository has been upgraded to Bazel 8, these files will be removed.

### Create MODULE.bazel

The new module declaration is organized into these sections:

1. Common (bazel_skylib)
2. Languages & toolchains: Python and Kotlin (required by the checkstyle rule from typedb_dependencies)
3. TypeDB repos (typedb_dependencies, typedb_bazel_distribution)


MODULE.bazel
```diff
@@ -0,0 +1,49 @@
...
+    remote = "https://github.com/typedb/dependencies",
+    commit = "ccdb36af9a868eeb6116d508db05a5b93e688e79",
+)
+
+bazel_dep(name = "typedb_bazel_distribution", version = "0.0.0")
+git_override(
+    module_name = "typedb_bazel_distribution",
+    remote = "https://github.com/typedb/bazel-distribution",
+    commit = "9796c9454b766c453456c8c0498dd67569db310e",
+)
```
krishnangovindraj: Do these need to be here or can the y be in repositories?

krishnangovindraj (approved):
Left a couple of questions.
Are we keeping both WORKSPACE and bazelmod in parallel?
Can we just get rid of WORKSPACE?

PR #405, opened 2026-03-17
Optional inserts receive concepts created in parent as "input"

krishnangovindraj:
## Usage and product changes
Adds a test checking the value of an attribute inserted in the root insert is available in try blocks in that insert.


dmitrii-ubskii (approved):

krishnangovindraj:
Looks like I missed merging this, but I've verified it passes passes in master.

PR #412, opened 2026-03-27
@doc and @meta annotations

dmitrii-ubskii:
## Product change and motivation

We implement `@doc("docstring")` and `@meta("key", "value")` schema annotations. These annotations can be attached to types and capabilities i.e. `owns`, `plays`, `relates`, and `sub`:

```php
define
entity person @doc("this represents an individual client") @meta("icon", ":silhouette.png"),
  owns name @doc("full name including title");
```
```php
match
  $x isa client;
  let $x_icon = get_meta("icon", $t); # get each client with a UI display icon hint
```

These annotations can also be attached to function definitions. They go at the end of the signature, before the colon introducing the body of the function:
```php
define
fun get_random_number() -> integer
    @doc("chosen by a fair dice roll"):
match
  let $rand = 4;
return first $rand;
```

## Implementation


query/language/annotations.feature
```diff
@@ -0,0 +1,386 @@
...
+        match
+          let $rand = 4;
+        return first $rand;
+      """
+    Then transaction commits
+    Then connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        let $metadata = function_get_doc("get_random_number");
```
flyingsilverfin: mistake

query/language/annotations.feature
```diff
@@ -0,0 +1,386 @@
...
+      define
+        relation person, <constraint> <rhs> @meta("key", "lorem ipsum");
+        relation base, relates arole;
+        attribute id value string;
+      """
+    Then transaction commits
+    Then connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match let $metadata = get_constraint_metadata("key", "person", "<constraint>", "<rhs>");
```
flyingsilverfin: I think this should be `get_xxx_meta` to match the annotation

query/language/annotations.feature
```diff
@@ -0,0 +1,386 @@
...
+  Scenario: retrieving a missing metadata annotation returns an empty string
+    When typeql schema query
+      """
+      define entity person;
+      """
+    Then transaction commits
+    Then connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match let $metadata = get_metadata("key", "person");
```
flyingsilverfin: Also I think the first arg should be the type, then the key? any specific reason it's this way around?

query/language/annotations.feature
```diff
@@ -0,0 +1,386 @@
...
+  Scenario: entity types can have doc annotations
+    When typeql schema query
+      """
+      define entity person @doc("This represents a person");
+      """
+    Then transaction commits
+    Then connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match let $doc = get_doc("person");
```
flyingsilverfin: I actually think that it's powerful to make this a type, not a string input, and also to allow a type variable to be passed in. Then users can do this:
```
match
  $t sub legal-entity;
fetch {
  "label": label($t),
  "doc": get_doc($t),
};
```

query/language/metadata.feature
```diff
@@ -0,0 +1,386 @@
...
+        match
+          let $rand = 4;
+        return first $rand;
+      """
+    Then transaction commits
+    Then connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        let $doc = get_function_doc("get_random_number");
```
flyingsilverfin: maybe `get_fun_doc`? to match keywords

query/language/annotations.feature
```diff
@@ -0,0 +1,386 @@
...
+      define
+        relation person, <constraint> <rhs> @doc("lorem ipsum");
+        relation base, relates arole;
+        attribute id value string;
+      """
+    Then transaction commits
+    Then connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match let $doc = get_constraint_doc("person", "<constraint>", "<rhs>");
```
flyingsilverfin: I thin it'll be much easier for people if we just have more functions:
- get_owns_doc
- get_plays_doc
- get_relates_doc

also shorter, and easier to use IMO:
```
match let $doc = get_owns_doc(person, name);
```

query/language/annotations.feature
dmitrii-ubskii: It feels like the key is the closest conceptually to the function name. `@meta("key", ...)` -> `get_meta("key", ...)`. That is, we're retrieving the `meta("key"` in a way.

flyingsilverfin (approved):

PR #415, opened 2026-04-16
Fix datetime format

dmitrii-ubskii:
## Usage and product changes

`chrono` updated its parsing from 0.4.41 to 0.4.44 to require exactly 9 digits in the nanosecond part, which broke our tests. This PR updates the offending values.


krishnangovindraj (approved):
wild

PR #418, opened 2026-04-23
Add tests for constraints producing the same variable

krishnangovindraj:
## Usage and product changes
Add tests for constraints producing the same variable twice.

### DIFF
--- a/query/functions/usage.feature
+++ b/query/functions/usage.feature
@@ -128,6 +128,18 @@ Feature: Function Usage
       let $five_again = 5;
     return first $five, $five_again;
 
+    fun five_stream() -> { integer }:
+    match
+      let $five = 5;
+    return { $five };
+
+    fun five_five_stream() -> { integer, integer }:
+    match
+      let $five = 5;
+      let $five_again = 5;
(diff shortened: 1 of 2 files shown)


query/functions/usage.feature
```diff
@@ -121,15 +121,51 @@ Feature: Function Usage
...
     """
     match
       let $five = five();
       let $five = five();
     """
+    Then typeql read query; fails with a message containing: "Variable 'five' cannot be assigned to multiple times in the same branch"
+    """
+    match
+      let $five, $five = five_five();
+    """
```
krishnangovindraj, line 156: This is existing behaviour

dmitrii-ubskii (commented):
Worth adding:
- three-way `$r($r, $r)`
- relation-player with another player: `$r($r, $x)`
- role types: `$r($t: $x, $t: $y)`
- maybe also a full-on `$r($t: $r, $t: $r)`, though this would have to produce empty results for now, with a TODO
- types: `$t sub $t`

flyingsilverfin (approved):
Nice!

PR #421, opened 2026-05-20
Add tests for given stage

krishnangovindraj:
## Usage and product changes
Add tests for given stage

### DIFF
--- a/driver/query.feature
+++ b/driver/query.feature
@@ -1015,7 +1015,7 @@ Feature: Driver Query
     """
 
 
-  Scenario: Driver processes query inputs correctly
+  Scenario: Driver processes query given rows correctly
     Given connection open schema transaction for database: typedb
     Given typeql schema query
       """
@@ -1035,14 +1035,14 @@ Feature: Driver Query
     Given transaction commits
 
     Given connection open write transaction for database: typedb
-    Given query inputs
+    Given query is given rows
       | p: person                    | age_value: integer? |
       | iid:0x1e00000000000000000001 | value:integer:23    |
       | iid:0x1e00000000000000000000 | none                |
 
     When get answers of typeql write query
     """
-    inputs $p: person, $age_value: integer;
+    given $p: person, $age_value: integer;
     match $p isa person, has name $name;
     insert try { $p has age == $age_value; };
     """
--- a/driver/query.feature
+++ b/driver/query.feature
@@ -1082,6 +1082,95 @@ Feature: Driver Query
     Then answer get row(0) get attribute(name) get value is: "Jane"
 
 
+  Scenario: The order of variables in given rows don't matter
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+      """
+      define
+        entity person owns name @card(0..), owns age @card(0..);
+        attribute name, value string;
+        attribute age, value integer;
+      """
+    Given transaction commits
+
+    Given connection open write transaction for database: typedb
+    Given get answers of typeql write query
+      """
+      insert $_ isa person, has name "John";
+      insert $_ isa person, has name "Jane";
+      """
+    Given transaction commits
+
+    Given connection open write transaction for database: typedb
+    Given set query option include_instance_types to: true
+    Given set answers of typeql read query as given rows with order: $p, $age_value
+    """
+    match
+      $p isa person, has name $name;
+      $name == "Jane"; try { let $age_value = 38; };
+    sort $name;
+    select $p, $age_value;
+    """
+    When get answers of typeql write query with given rows
+    """
+    given $p: person, $age_value: integer?;
+    match $p isa person, has name $name;
+    insert try { $p has age == $age_value; };
+    """
+    # We don't necessarily guarantee that the rows retain the order
+    Then answer get row(0) get concepts size is: 3
+
+    Then answer get row(0) get entity(p) get type get label: person
+    Then answer get row(0) get attribute(name) get type get label: name
+    Then answer get row(0) get attribute(name) get value is: "Jane"
+    Given set answers of typeql read query as given rows with order: $p
+    """
+    match
+      $p isa person, has name $name;
+      $name == "John";
+    sort $name;
+    select $p;
+    """
+    When get answers of typeql write query with given rows
+    """
+    given $p: person, $age_value: integer?;
+    match $p isa person, has name $name;
+    insert try { $p has age == $age_value; };
+    """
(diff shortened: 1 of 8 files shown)


driver/query.feature
```diff
@@ -1015,6 +1015,77 @@ Feature: Driver Query
...
+    match
+      $p isa person, has name $name;
+      {
+        $name == "Jane"; try { let $age_value = 38; };
+      } or {
+        $name == "John"; try { let $age_value = 0; $age_value == 1; };
+      };
+    sort $name;
+    select $p, $age_value;
+    """
```
flyingsilverfin, line 1050: I see why would do this so you can test with Concepts as well haha ok fun

driver/query.feature
```diff
@@ -1015,6 +1015,77 @@ Feature: Driver Query
...
+    When get answers of typeql write query with given rows
+    """
+    given $p: person, $age_value: integer?;
+    match $p isa person, has name $name;
+    insert try { $p has age == $age_value; };
+    """
+    # We don't necessarily guarantee that the rows retain the order
```
flyingsilverfin, line 1057: Is this a TODO?

flyingsilverfin: We can probably just assert the size to avoid this issue and check the values in the follow up query

query/language/given.feature
```diff
@@ -0,0 +1,285 @@
...
+      insert $_ isa person, has name "John", has age 25, has ref 100;
+      insert $_ isa person, has name "Jane", has age 30, has ref 101;
+      insert $_ isa company, has name "TypeDB", has ref 200;
+      """
+    Given transaction commits
+
+
+    Scenario: raw values can be used in given rows
+      Given connection open read transaction for database: typedb
+      Given query is given rows
```
flyingsilverfin: Cool this is the other way, nice

query/language/given.feature
```diff
@@ -0,0 +1,285 @@
...
+
+
+  Scenario: Given rows are checked against declared types
+    Given connection open read transaction for database: typedb
+    # Values: Pass a string instead
+    Given query is given rows
+      | x                |
+      | value:integer:3  |
+      | value:string:abc |
+    Then typeql read query with given rows; fails with a message containing: "The given value at row '1' and column '0' does not not satisfy the declared type"
```
flyingsilverfin: nice!

query/language/given.feature
```diff
@@ -0,0 +1,285 @@
...
+      select $person;
+      """
+    Then uniquely identify answer concepts
+      | person      |
+      | key:ref:100 |
+
+    Given query is given rows
+      | person           |
+      | iid:entity:0:123 |
+    Then typeql read query with given rows; fails with a message containing: "The given instance at row '0' and column '0' does not exist in the database"
```
flyingsilverfin, line 238: I wonder if we should treat this like looking up `$x iid <iid>;` and just throw the row out if it doesn't exist?

query/language/given.feature
```diff
@@ -0,0 +1,285 @@
...
+        match $p isa person, has name $name;
+        """
+      Then uniquely identify answer concepts
+        | p           | name            |
+        | key:ref:100 | attr:name:John  |
+        | key:ref:101 | attr:name:Jane  |
+        | key:ref:110 | attr:name:James |
+
+
+  Scenario: concepts given can be subtypes of the declared types
```
flyingsilverfin, line 269: nice

query/language/given.feature
```diff
@@ -0,0 +1,285 @@
...
+      match $p isa person, has name $name, has weight $weight;
+      """
+    Then uniquely identify answer concepts
+      | p           | name           | weight        |
+      | key:ref:101 |attr:name:Jane | attr:weight:65 |
+    Given transaction commits
+
+    # Bonus, ensure the bounds are tight.
+    Given connection open write transaction for database: typedb
+    Given query is given rows
```
flyingsilverfin, line 300: What happens if we pass no rows? Will this error still happen?

query/language/given.feature
```diff
@@ -0,0 +1,285 @@
...
+        insert $p isa person, has ref == $ref, has name == $name;
+        """
+
+    # Declared None, used outside try
+    Given connection open write transaction for database: typedb
+    Given query is given rows
+      | ref               | name               |
+      | value:integer:110 | value:string:James |
+      | value:integer:111 | none               |
+    Then typeql write query with given rows; fails with a message containing: "A write stage uses the optional variable 'name' outside a 'try' block"
```
flyingsilverfin, line 330: For reference: when we sort out optionals, we will actually allow this but if you don't use `try` it will fail the entire insert if any of the variables using `!` are optional and not in a `try` right?

flyingsilverfin (approved):
Great stuff! A couple of comments to implement

query/language/given.feature
flyingsilverfin: Probably a good test to have actually as well - 0 input rows

query/language/given.feature
krishnangovindraj: No, that would still be a compilation error.

driver/query.feature
krishnangovindraj: > Is this a TODO?
No, it's just not in our semantics that we guarantee the order. I could add a sort but I imagine all of these tests have the same problem and are happy making an assumption

query/language/given.feature
krishnangovindraj: I think I'd rather have a strong error.

query/language/given.feature
krishnangovindraj: Yes, the error should still happen.
Added a separate test for 0 rows which just returns 0 rows.

query/language/given.feature
flyingsilverfin: What's the motivation? Looking up an IID that doesn't exist in any other part of the language isn't an error right?

query/language/given.feature
flyingsilverfin: wait you won't be able to pass a `None` into a non-optional variable declaration right

query/language/given.feature
krishnangovindraj: It's an interface between the language and the outside world, so it's ok to have a strong error here. Although we could move the check to the decode stage instead of the execution phase if needed. 
We could in fact get rid of the whole given stage at this point and move it all back into decode and keep it as a precondition.

krishnangovindraj: But also, pattern-matching constraints fail. This isn't a constraint.

PR #423, opened 2026-06-18
Test UTF-8 correctness in driver

dmitrii-ubskii:
## Usage and product changes

We add test cases containing characters with various UTF-8 encoding lengths.


farost (changes requested):
There is one `@ignore-typedb-driver-java` in `database.feature` with a `TODO` and the same `@ignore-typedb-driver-java` in `user.feature`.  Let's cover these here.

farost (approved):

PR #424, opened 2026-06-22
Allow committing schema functions on abstract types

krishnangovindraj:
## Usage and product changes
Add tests which allow schema functions which satisfy type-checking only through abstract types


dmitrii-ubskii (approved):
Some of the `Given`s should be a `Then`.

flyingsilverfin (approved):

PR #425, opened 2026-06-30
Test doc and metadata functions in fetch

dmitrii-ubskii:
## Usage and product changes

We add tests for built-in documentation and metadata retrieval function usage in fetch stages. This is following a bug where built-in function name resolution was not performed correctly in `fetch`.

### DIFF
--- a/query/language/metadata.feature
+++ b/query/language/metadata.feature
@@ -26,6 +26,18 @@ Feature: TypeQL schema metadata
     Then uniquely identify answer concepts
       | doc                                   |
       | value:string:This represents a person |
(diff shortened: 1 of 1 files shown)


farost (approved):
We usually put all fetch tests into `fetch.feature`, but I guess
* this file has good coverage with the common setup between the regular matches and fetches
* we don't really do anything very special in terms of the syntax and "type" of the expressions/return types, so it's technically already covered in `fetch.feature`.

If there is a fetch-related bug to cover with tests, making a small test in `fetch.feature` could be good.

PR #427, opened 2026-07-10
String attribute comparison tests

dmitrii-ubskii:
## Usage and product changes

We add tests for comparing strings of various sizes to catch encoding level bugs.

### DIFF
--- a/query/language/expression.feature
+++ b/query/language/expression.feature
@@ -1316,11 +1316,16 @@ Feature: TypeQL Query with Expressions
       | value:string:rel:rolename |
 
 
-  Scenario: String attributes are retrieved correctly when compared across lengths
-    Given connection open write transaction for database: typedb
-    Given typeql write query
-    """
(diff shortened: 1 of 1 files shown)


query/language/expression.feature
```diff
@@ -1313,3 +1314,487 @@ Feature: TypeQL Query with Expressions
...
+            $len > 1;
+            let $a_ in names_helper($len - 1);
+            let $a = "a" + $a_;
+          } or {
+            let $a = "a";
+          };
+        return { $a };
+      with fun names() -> { string }:
+        match let $a in names_helper(20);
+        return { $a };
```
dmitrii-ubskii: Note, this only works because we have increased stack size in our BDD tests.

query/language/expression.feature
krishnangovindraj, line 1: Do you think we need a few tests comparing against a raw value?

krishnangovindraj (approved):

PR #429, opened 2026-07-10
Add BDD scenarios for 'vector' value type

lolski:
## Usage and product changes

Adds BDD scenarios for 'vector' value type and cosine-similarity vector search operation.

## Implementation

- Added defining double array scenarios in `define.feature`
- Added inserting double array scenarios in `insert.feature`
- Added match scenario in `match.feature`

### DIFF
--- a/query/language/define.feature
+++ b/query/language/define.feature
@@ -1083,7 +1083,7 @@ Feature: TypeQL Define Query
   Scenario: an attribute type can be defined with a fixed-size vector value type
     When typeql schema query
       """
-      define attribute embedding value vector(3, f8);
+      define attribute embedding value vector(3, "float32");
       """
     Then transaction commits
 
@@ -1092,7 +1092,7 @@ Feature: TypeQL Define Query
     When typeql schema query
       """
       define
-      attribute embedding value vector(3, f8);
+      attribute embedding value vector(3, "float32");
       entity document owns embedding;
       """
     Then transaction commits
@@ -1110,7 +1110,7 @@ Feature: TypeQL Define Query
   Scenario: defining an embedding value type with a non-integer length errors
     Then typeql schema query; fails
       """
-      define attribute embedding value vector(abc, f8);
+      define attribute embedding value vector(abc, "float32");
       """
 
 
--- a/query/language/define.feature
+++ b/query/language/define.feature
@@ -1110,7 +1110,7 @@ Feature: TypeQL Define Query
   Scenario: defining an embedding value type with a non-integer length errors
     Then typeql schema query; fails
       """
-      define attribute embedding value vector(abc, "float32");
+      define attribute embedding value vector(3.0, "float32");
       """
 
 
--- a/query/language/define.feature
+++ b/query/language/define.feature
@@ -1121,6 +1121,13 @@ Feature: TypeQL Define Query
       """
 
 
+  Scenario: defining an embedding value type with a negative length errors
+    Then typeql schema query; fails
+      """
+      define attribute embedding value vector(-3, "float32");
+      """
+
+
   Scenario: new attribute type declaration should contain kind
     Then typeql schema query; fails
       """
--- a/query/language/define.feature
+++ b/query/language/define.feature
@@ -1080,6 +1080,49 @@ Feature: TypeQL Define Query
       | duration    | procedure-duration |
 
 
+  Scenario: an attribute type can be defined with a fixed-size double-array value type
+    When typeql schema query
+      """
+      define attribute embedding value double[3];
+      """
+    Then transaction commits
+
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        $x label embedding;
+        attribute $x;
+      """
+    Then answer size is: 1
+
+
+  Scenario: an entity type can own an embedding attribute type
+    When typeql schema query
+      """
+      define
+      attribute embedding value double[3];
+      entity document owns embedding;
+      """
+    Then transaction commits
+
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        $x label document, owns $e;
+        $e label embedding;
+      """
+    Then answer size is: 1
+
+
+  Scenario: defining an embedding value type with a non-integer length errors
+    Then typeql schema query; fails
+      """
+      define attribute embedding value double[abc];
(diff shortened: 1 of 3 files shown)


query/language/define.feature
```diff
@@ -1080,6 +1080,49 @@ Feature: TypeQL Define Query
       | duration    | procedure-duration |
 
 
+  Scenario: an attribute type can be defined with a fixed-size double-array value type
+    When typeql schema query
+      """
+      define attribute embedding value double[3];
```
dmitrii-ubskii: My biggest problem with this syntax is that it implies to me, as a user, that `integer[3]` and `string[3]` should also be allowed. I'd much prefer special syntax like `vector(3)` or `vector<3>. It has less danger of being confused with general purpose arrays, _and_ hints at vector indexing support.

query/language/define.feature
```diff
@@ -1080,6 +1080,49 @@ Feature: TypeQL Define Query
...
+    Then transaction commits
+
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        $x label embedding;
+        attribute $x;
+      """
+    Then answer size is: 1
```
dmitrii-ubskii: Seems redundant.

query/language/insert.feature
```diff
@@ -2894,3 +2894,50 @@ Parker";
...
+      entity document owns embedding;
+      """
+    Given transaction commits
+
+    Given connection open write transaction for database: typedb
+    When typeql write query
+      """
+      insert
+      $x isa document,
+        has embedding [0.1, 0.2, 0.3];
```
dmitrii-ubskii: Is a vector expected to be interchangeable with a list of doubles? I.e. would it support indexing with `$emb[2]` e.g.?

query/language/match.feature
```diff
@@ -5630,3 +5630,39 @@ Feature: TypeQL Match Clause
...
+      """
+    Then transaction commits
+
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        $doc isa document, has name $name, has embedding $e;
+        let $score = cosine_similarity($e, [1.0, 0.0, 0.0]);
+        $score >= 0.8;
```
dmitrii-ubskii: We're not providing vector indexing? This looks like it would check the embedding of every name in the database which sort of defeats the point.

I'd expect something like:
```
match
  let $emb in cosine_similarity_search([1.0, 0.0, 0.0], 0.8);
  $doc isa document, has name $name, has embedding $emb;
```

I'm also not super comfortable with just a bare vector value here rather than an attribute. Surely we're not going to have an index spanning _all_ the 3-dimensional vectors? Same length vectors can come from different models entirely.

dmitrii-ubskii (changes requested):

query/language/define.feature
krishnangovindraj: Or in analogy to structs. 
```
vector Qwen3-4B-Q4_K_M<2560,q4_k_m>;
```
This strongly types the vectors so the type-system can error if they come from different models. But I'll admit that's very overkill, so I'd just go for Dmitrii's, with the encoding if it's relevant.

query/language/define.feature
```diff
@@ -1080,6 +1080,40 @@ Feature: TypeQL Define Query
...
+        $x label document, owns $e;
+        $e label embedding;
+      """
+    Then answer size is: 1
+
+
+  Scenario: defining an embedding value type with a non-integer length errors
+    Then typeql schema query; fails
+      """
+      define attribute embedding value vector(abc, "float32");
```
dmitrii-ubskii: It should at least be a value for the test to make sense
```suggestion
      define attribute embedding value vector(3.0, "float32");
```

query/language/insert.feature
```diff
@@ -2894,3 +2894,50 @@ Parker";
...
+    When connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        $x isa document, has embedding $e;
+      """
+    Then answer size is: 1
+
+
+  Scenario: inserting an embedding whose length differs from the declared array size errors
```
dmitrii-ubskii, line 2927: Needs a variant with encoding mismatch too

query/language/match.feature
```diff
@@ -5630,3 +5630,37 @@ Feature: TypeQL Match Clause
...
+      $near isa document, has name "near", has embedding vector([1.0, 0.0, 0.0], "float32");
+      $far isa document, has name "far", has embedding vector([0.0, 1.0, 0.0], "float32");
+      """
+    Then transaction commits
+
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        let $e in cosine_similarity_search(embedding, vector([1.0, 0.0, 0.0], "float32"), 0.8);
```
dmitrii-ubskii, line 5657: Nice.

dmitrii-ubskii (approved):

query/language/define.feature
lolski: how many non-happy path scenario should we test in this case you think? would it be overkill if we test double, string, and something like `vector(abc, "float32");`? seems like a pretty trivial case imo and we don't need to invest too much in adding too many cases

query/language/define.feature
lolski: Resolved in person, we're settling with the following syntax: `vector(<num-of-elems>, "<datatype>")`

query/language/define.feature
lolski: Fixed

query/language/insert.feature
lolski: Resolved in person, we're settling with the following syntax: `vector([elem1, elem2, ...], "<datatype>")` for defining literal attribute

query/language/match.feature
lolski: Fixed

query/language/define.feature
lolski: Fixed according to your suggestion. I also added another case where you shouldn't allow negative number eg., `vector(-3, "float")`.

query/language/insert.feature
lolski: Skipped for now as we're starting with only one encoding: 32-bit float. We'll introduce this case when we add other types such as when 8-bit float (most likely when we want to add quantisation feature)

PR #430, opened 2026-07-19
Extend given-rows scenarios

flyingsilverfin:
## Usage and product changes

Extend the given-rows test suite to cover all value types and the raw HTTP api, including ones containing quotes in them that have to deal with escaping.

## Implementation

- Add new HTTP-api specific test for Given rows across all value types, testing raw (eg. just a string) and structured values (eg, a Value concept containing a string)
- Add driver (gRPC) query API tests for all Given row value types
- Add server-side embedded Given row tests for all value types

### DIFF
--- a/driver/http/http.feature
+++ b/driver/http/http.feature
@@ -1084,3 +1084,57 @@ Feature: TypeDB HTTP Endpoint
       }
     }
     """
+
+  ##############
+  # GIVEN ROWS #
+  ##############
+
+  Scenario Outline: Given rows handle <encoding> <value-type> values
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+      """
+      define entity owner, owns attr; attribute attr, value <value-type>;
+      """
+    Given transaction commits
+
+    Given connection open write transaction for database: typedb
+    When set query given rows
+      | attr                            |
+      | <encoding>:<value-type>:<value> |
+    When get answers of typeql write query with given rows
+      """
+      given $attr: <value-type>;
+      insert $x isa owner, has attr == $attr;
+      """
+    Given transaction commits
+
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match $x isa owner, has attr $attr;
+      """
+    Then answer size is: 1
+    Then answer get row(0) get attribute(attr) get value is: <value>
+
+    Examples:
+      | encoding | value-type  | value                              |
+      | raw      | boolean     | true                               |
+      | raw      | integer     | 12345090                           |
+      | raw      | double      | 2.01234567                         |
+      | raw      | decimal     | 1234567890.0001234567890dec        |
+      | raw      | date        | 2024-09-20                         |
+      | raw      | datetime    | 1999-02-26T12:15:05                |
+      | raw      | datetime-tz | 2024-09-20T16:40:05.000000001+0100 |
+      | raw      | duration    | P1Y10M7DT15H44M5.00394892S         |
+      | raw      | string      | hello                              |
+      | raw      | string      | bob"bobby                          |
+      | value    | boolean     | true                               |
+      | value    | integer     | 12345090                           |
+      | value    | double      | 2.01234567                         |
+      | value    | decimal     | 1234567890.0001234567890dec        |
+      | value    | date        | 2024-09-20                         |
+      | value    | datetime    | 1999-02-26T12:15:05                |
+      | value    | datetime-tz | 2024-09-20T16:40:05.000000001+0100 |
+      | value    | duration    | P1Y10M7DT15H44M5.00394892S         |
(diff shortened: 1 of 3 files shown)


driver/http/http.feature
```diff
@@ -1084,3 +1084,98 @@ Feature: TypeDB HTTP Endpoint
...
+          owns is-active, owns item-count, owns average-rating, owns account-balance, owns birth-date,
+          owns created-time, owns event-time, owns time-elapsed, owns full-name;
+        entity value-record,
+          owns is-active, owns item-count, owns average-rating, owns account-balance, owns birth-date,
+          owns created-time, owns event-time, owns time-elapsed, owns full-name;
+      """
+    Given transaction commits
+
+    Given connection open write transaction for database: typedb
+    When set given rows to:
```
flyingsilverfin: Set explicitly by "Concept::Value" or by raw value, to fully test the HTTP api (this is where a bug had been found!)

flyingsilverfin:
Interesting to me that we now have 2 places that have basically duplicate test: in driver.feature and in the language/<test>... but that's because we don't have all the server tests running in driver (good). One idea could I suppose be to have the driver tests run against embedded typedb too?

For now ok with the duplication, the intent remains, in practical terms we have 4 domains
1) server behaviour tests (runs server-side), which ensures the details and behaviours of the server
2) language behaviour tests (run server-side), which ensures the details and the behaviours of the language
3) driver tests (for driver libraries, such as gRPC ones or http-ts which have an API to test), which ensure the APIs we present work
4) http tests (test the raw HTTP API), which ensures the API we present works

driver/query.feature
```diff
@@ -1266,6 +1266,90 @@ Feature: Driver Query
     Then answer get row(0) get variable(y) is empty
     Then answer get row(0) get variable(q) is empty
 
+
+  Scenario: Given rows handle all value types
```
flyingsilverfin: New test that does all value types in one go

query/language/given.feature
```diff
@@ -67,6 +67,80 @@ Feature: TypeQL Given Clause
       | key:ref:101 |
 
 
+  Scenario: Given rows handle all value types
```
flyingsilverfin: Test all value types server side

krishnangovindraj: Would be cleaner as a scenario outline.

driver/query.feature
krishnangovindraj: We had an integration test per language, but this is good too depending on how it's implemented in BDD

driver/http/http.feature
krishnangovindraj: Cleaner as a scenario outline.
That would also let us add extra cases like escaped unicode characters etc.

krishnangovindraj (changes requested):
Good tests to have, better to have as scenario outlines though.

driver/http/http.feature
```diff
@@ -1084,3 +1084,57 @@ Feature: TypeDB HTTP Endpoint
...
+      | value    | boolean     | true                               | true                               |
+      | value    | integer     | 12345090                           | 12345090                           |
+      | value    | double      | 2.01234567                         | 2.01234567                         |
+      | value    | decimal     | 1234567890.0001234567890           | 1234567890.0001234567890dec        |
+      | value    | date        | 2024-09-20                         | 2024-09-20                         |
+      | value    | datetime    | 1999-02-26T12:15:05                | 1999-02-26T12:15:05                |
+      | value    | datetime-tz | 2024-09-20T16:40:05.000000001+0100 | 2024-09-20T16:40:05.000000001+0100 |
+      | value    | duration    | P1Y10M7DT15H44M5.00394892S         | P1Y10M7DT15H44M5.00394892S         |
+      | value    | string      | hello                              | "hello"                            |
+      | value    | string      | bob"bobby                          | "bob\"bobby"                       |
```
krishnangovindraj: Do we need duplicate columns? Is that for the quotes? 
This is a bit weird because it's dependent on the step implementation, isn't it?
In the integration test, I was testing it against the value encoded into the query.

driver/http/http.feature
```diff
@@ -1084,3 +1084,57 @@ Feature: TypeDB HTTP Endpoint
...
+  Scenario Outline: Given rows handle <encoding> <value-type> values
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+      """
+      define entity owner, owns attr; attribute attr, value <value-type>;
+      """
+    Given transaction commits
+
+    Given connection open write transaction for database: typedb
+    When set given rows to:
```
krishnangovindraj: Also, We use a "query is given rows" step in the core BDD (also in this PR in given.feature)

krishnangovindraj (approved):
Just one comment about "set given rows to:" and "query is given rows"

PR #431, opened 2026-07-29
Add tests for renaming types

krishnangovindraj:
## Usage and product changes
Add tests for renaming entity, relation, attribute& role types.

### DIFF
--- a/query/functions/definition.feature
+++ b/query/functions/definition.feature
@@ -760,6 +760,13 @@ Feature: Function Definition
     nickname;
     """
     Then transaction commits; fails
+
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+    """
+    redefine attribute nickname label friendly-alias;
+    """
+    Then transaction commits; fails
     # TODO: Add messsage when it's an explicit check at query time
 
 
--- a/query/functions/definition.feature
+++ b/query/functions/definition.feature
@@ -760,6 +760,7 @@ Feature: Function Definition
     nickname;
     """
     Then transaction commits; fails
+    # TODO: Add messsage when it's an explicit check at query time
 
     Given connection open schema transaction for database: typedb
     Given typeql schema query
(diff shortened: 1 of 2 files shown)


query/functions/definition.feature
```diff
@@ -760,6 +760,13 @@ Feature: Function Definition
     nickname;
     """
     Then transaction commits; fails
+
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+    """
+    redefine attribute nickname label friendly-alias;
+    """
+    Then transaction commits; fails
```
krishnangovindraj, line 770: Not essential behaviour, but just a side-effect of storing functions as strings that I wanted to make sure failed.
The fact that this is at commit-time helps, else it would be a lot less straightforward.

flyingsilverfin (approved):
Approved but needs 'collision' tests with the difficulties of cross hierarchy, in-hierarchy, and overriding role types etc.

query/language/redefine.feature
```diff
@@ -368,6 +409,46 @@ Feature: TypeQL Redefine Query
...
+     entity cannot-rename-to;
+     relation parent-of-rename @abstract;
+     relation before-rename, relates dummy, sub parent-of-rename;
+     entity before-player plays before-rename:dummy;
+    """
+    Given transaction commits
+    Given connection open schema transaction for database: typedb
+    Then typeql schema query; fails with a message containing: "Label 'cannot-rename-to' should be unique, but is already used by 'entity'"
+      """
+      redefine entity before-rename label cannot-rename-to;
```
flyingsilverfin: wait this is renaming the entity still

query/language/redefine.feature
```diff
@@ -718,6 +847,48 @@ Feature: TypeQL Redefine Query
       attribute email value string @regex("^.*@.*$") @range("A".."zzzzzzzzzzzzzzzzzzzzzzzzzz");
       """
 
+
+  Scenario: attribute types can be renamed
```
flyingsilverfin, line 851: Can we add one more - commented out - struct rename (?) what do we think @dmitrii-ubskii  as well

PR #432, opened 2026-08-04
Extend database deletion and migration resilience scenarios

farost:
## Usage and product changes
Add new scenarios to ensure the delete and import operations survive a couple of previously uncovered, but important corner cases:
- Database can be deleted with the same name right after a transaction commit (verifying that there are no lock leaking)
- Concurrent database imports cannot get in the way of each other, and the data is not corrupted in case of incorrectly terminated import processes

### DIFF
--- a/connection/database.feature
+++ b/connection/database.feature
@@ -157,6 +157,21 @@ Feature: Connection Database
     Then connection delete database: typedb
     Then connection does not have database: typedb
 
+  Scenario: database can be deleted and recreated immediately after committing transactions
+    When connection create database: typedb
+    When connection open schema transaction for database: typedb
+    Then typeql schema query
+      """
+      define entity person;
+      """
+    Then transaction commits
+    When connection open write transaction for database: typedb
+    Then transaction commits
+    Then connection delete database: typedb
+    Then connection does not have database: typedb
+    When connection create database: typedb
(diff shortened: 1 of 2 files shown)


connection/database.feature
```diff
@@ -157,6 +157,21 @@ Feature: Connection Database
     Then connection delete database: typedb
     Then connection does not have database: typedb
 
+  Scenario: database can be deleted and recreated immediately after committing transactions
```
farost, line 160: Found a bug in typedb. Fixed in https://github.com/typedb/typedb/pull/7892, covered with a scenario

driver/migration.feature
```diff
@@ -1616,3 +1616,89 @@ Feature: Driver Migration
         $p isa person, has name "P4", has email $n;
       """
     Then answer size is: 0
+
+
+  @ignore-typedb-driver-java @ignore-typedb-driver-python @ignore-typedb-driver-csharp
```
farost, line 1621: Only the Rust driver has an option to manipulate the stream manually (although we don't want to expose it to the users, it's the only way we can currently test migration with more precision. I think it's valuable)

flyingsilverfin (approved):
Looks good!

dmitrii-ubskii (approved):

PR #433, opened 2026-08-05
Vector search

lolski:

### DIFF
--- a/query/language/match.feature
+++ b/query/language/match.feature
@@ -5632,8 +5632,6 @@ Feature: TypeQL Match Clause
       """
 
 
-  # TODO: enable when similarity search functions are implemented (deferred vector work)
-  @ignore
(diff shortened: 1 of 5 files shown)


query/language/match.feature
```diff
@@ -5666,3 +5664,245 @@ Feature: TypeQL Match Clause
...
+      $a isa document, has name "a", has embedding vector([1.0, 0.0, 0.0], "float32");
+      $b isa document, has name "b", has embedding vector([1.0, 0.0, 0.0], "float32");
+      """
+    Then transaction commits
+
+    Given connection open read transaction for database: typedb
+    When get answers of typeql read query
+      """
+      match
+        let $e in cosine_similarity_search(embedding, vector([1.0, 0.0, 0.0], "float32"), 0.8);
```
flyingsilverfin, line 5899: On top of providing this function, do we want to also provide a function `cosine_similarity` to actually compute the value?

PR #434, opened 2026-08-09
Add tests for whether conjunctions can have valid plans

krishnangovindraj:
## Usage and product changes
Add tests for whether there can exist a valid plan for conjunctions.

### DIFF
--- a/query/functions/usage.feature
+++ b/query/functions/usage.feature
@@ -248,8 +248,8 @@ Feature: Function Usage
 
   Scenario: When no evaluation order can bind the arguments of a function, an error is returned.
     Given connection open read transaction for database: typedb
-    # "The conjunction cannot have a valid plan" is also fine here, but we have easier pre-checks at the moment.
-    Then typeql read query; fails with a message containing: "The variable 'x' is required to be bound to a value before it's used."
+    # "The required input variables for the following constraints could not be satisfied (there may be a circular dependency)" is also fine here, but we have easier pre-checks at the moment.
+    Then typeql read query; fails with a message containing: "The variable 'x' must be bound to a value before it's used."
     """
       with fun plus($x: integer, $inc: integer) -> integer:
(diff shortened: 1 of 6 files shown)


query/language/expression.feature
```diff
@@ -171,6 +214,22 @@ Feature: TypeQL Query with Expressions
       select
         $x, $u, $v;
       """
+    Given transaction closes
+
+    # Non-cyclic query, just inverted dependencies in two branches.
+    # TODO: This is legal, but currently fails because our validation is too coarse.
+    # The validation introduced in #7900 is finer and permits this.
```
krishnangovindraj, line 221: This is set up to break when we remove the old coarse validation.

query/functions/usage.feature
```diff
@@ -245,3 +245,31 @@ Feature: Function Usage
...
+      match let $y = $x + $inc;
+      return first $x;
+
+      match
+        let $x = plus($x, 0);
+      """
+    Given transaction closes
+
+    Given connection open read transaction for database: typedb
+    Then typeql read query; fails with a message containing: "The conjunction cannot have a valid plan"
```
dmitrii-ubskii: "plan" isn't useful to the user in an error message. Can we say something like "circular dependency detected" instead?

query/language/expression.feature
```diff
@@ -147,11 +145,56 @@ Feature: TypeQL Query with Expressions
       select
         $x, $v;
       """
+    Given transaction closes
+    Given connection open read transaction for database: typedb
+    Then typeql read query; fails with a message containing: "The variable 'v' is required to be bound to a value before it's used."
```
dmitrii-ubskii:
```suggestion
    Then typeql read query; fails with a message containing: "The variable 'v' must be bound to a value before it's used."
```

dmitrii-ubskii (approved):

PR #435, opened 2026-08-13
Add scenarios for import with redeclared annotations

farost:
## Usage and product changes
Add `driver/migration` scenarios for bugs fixed in https://github.com/typedb/typedb/pull/7912:
1. `Independent sub attributes`: verify that `@independent` annotations declared on sub attributes don't block database import.
2. `Ownerships and roleplaying specializations`: verify that all main annotations defined on a self-specialization of an interface type don't block and survive database import.

### DIFF
--- a/driver/migration.feature
+++ b/driver/migration.feature
@@ -1150,6 +1150,233 @@ Feature: Driver Migration
     Then answer size is: 3
 
 
+  Scenario: Export and import of a database with attribute hierarchies declared in either order succeeds
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+      """
+      define
+        attribute technical-fit sub assessment-score, value integer;
+        attribute assessment-score @abstract;
+        attribute leaf-score @independent, sub rating, value integer;
+        attribute rating @abstract;
+        attribute measure @abstract;
+        attribute cpu-load sub measure, value integer;
+        attribute quality @abstract;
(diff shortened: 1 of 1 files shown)


driver/migration.feature
```diff
@@ -1150,6 +1150,233 @@ Feature: Driver Migration
...
+    Then answer size is: 2
+
+
+  Scenario: Export and import of a database with specializing owns keys and annotations succeeds
+    Given connection open schema transaction for database: typedb
+    Given typeql schema query
+      """
+      define
+        entity page @abstract, owns ref, owns id @unique, owns tag, owns kind @values(1, 2, 3), owns score @range(0..100), owns handle, plays team:member @doc("who can join");
+        entity profile sub page, owns ref @key, owns id @card(0..1) @unique, owns tag @card(0..2) @values("alpha", "beta"), owns kind @card(0..1) @values(1, 2), owns score @card(0..1) @range(10..50), owns handle @card(0..1) @regex("[a-z]+"), plays team:member @card(0..1) @doc("who can join");
```
dmitrii-ubskii, line 1255: Is it worth testing the other order of these two?

dmitrii-ubskii (approved):

driver/migration.feature
farost: I don't think so

PR #438, opened 2026-08-20
Update circular-assignments test for improved resolution

krishnangovindraj:
Parked so I don't forget it. The new way of doing it let's us do this.


krishnangovindraj:
Absorbed

PR #439, opened 2026-08-28
Optional safety tests but with transition version behaviour

krishnangovindraj:


query/language/fetch.feature
krishnangovindraj, line 1: Please see diff

PR #446, opened 2026-09-14
Extend fetch query analyze tests with attribute attribute

flyingsilverfin:
## Usage and product changes

Add a case to test the fetching of an attribute kind in analyze queries.

## Implementation

* Add attribute type to analyze query


krishnangovindraj (approved):
You could add relation too if you wanted to.

PR #448, opened 2026-09-16
Adds tests for optional expressions & short-circuits

krishnangovindraj:
## Usage and product changes
Adds tests for optional expressions & short-circuits


query/language/variables.feature
```diff
@@ -292,3 +296,288 @@ Feature: TypeQL Variable binding tests
     match
       let $x = ident($x);
     """
+
```
krishnangovindraj, line 299: Please review

query/language/expression.feature
```diff
@@ -2412,3 +2410,85 @@ Feature: TypeQL Query with Expressions
       | function       | a       | b       | result_type | result  |
       | std::math::min | 10.2dec | 13.5dec | decimal     | 10.2dec |
       | std::math::max | 10.2dec | 13.5dec | decimal     | 13.5dec |
+
+
+  ###############
```
krishnangovindraj, line 2415: Please review

krishnangovindraj (commented):
It's mostly expression.feature and variables.feature that need review since the rest will be changed to `if` statements elsewhere in the stack

query/language/expression.feature
```diff
@@ -2410,3 +2410,85 @@ Feature: TypeQL Query with Expressions
...
+      let $z = $y;
+    """
+    When get answers of typeql read query
+    """
+    match
+      { try { let $x = 5; }; } or { try { let $x = 5; $x == 4; }; };
+    match
+      let $y? = $x;
+    match
+      let $z? = $y;
```
flyingsilverfin: This isn't what is expected. Assignment can handle optionality fine. You also don't need to declare the LHS as `?`. 

There's two uses of `?` only:
1) Type signature. In this case, here it would be:
```
let $y: integer? = $x;
```
if we ever wanted to allow making it explicit

2) short circuit
```
let $y = $x?;
```
Which isn't required in this case, but can be used.

We don't ever annotate the variable itself at declaration site. We only need to acknowledge the presence of optionality at _usage_ sites. This encompasses the entire design. 

Second line in https://github.com/typedb/typedb/issues/7879#issuecomment-5623863840

query/language/expression.feature
```diff
@@ -2410,3 +2410,85 @@ Feature: TypeQL Query with Expressions
...
+      | value:integer:5 | value:integer:5 | value:integer:5 |
+      | none            | none            | none            |
+
+
+  Scenario: Sub-expressions returning optional values can short-circuit the expression to return None using '?'
+    Given connection open read transaction for database: typedb
+    Then typeql read query; fails with a message containing: "The expression 'Variable(x)' returns an optional value which may be empty. Use '?' to short-circuit and assign an empty result"
+    """
+    match
+      { try { let $x = 5; }; } or { try { let $x = 5; $x == 4; }; };
```
flyingsilverfin, line 2459: hum i guess this is nice but why are we using such a complex `try/or` to generate an optional variable? Why not just do `match try { let $x = 5; }` - Isn't `$x` going to be optional from that on its own?

Same for the following and preceding scenarios

query/language/fetch.feature
```diff
@@ -880,14 +880,25 @@ Feature: TypeQL Fetch Query
 
 
   Scenario: fetch can handle optional objects
+    Then typeql read query; fails with a message containing: "The optional variable 'e' was used unsafely in a fetch statement"
```
flyingsilverfin: Can we improve the error message?
```
Optional variable 'e' cannot be used in a fetch entry without '?'. Use '?' to propagate the empty optional into the fetch entry.
```

query/language/fetch.feature
```diff
@@ -919,14 +930,34 @@ Feature: TypeQL Fetch Query
       }
       """
 
+    Then typeql read query; fails with a message containing: "The optional variable 'e' was used in a context where it may fail the branch if unset"
```
flyingsilverfin: Can this be `Optional variable 'e' cannot be used without handling the empty value. Use 'isset $e' to require non-empty values.`
?

flyingsilverfin (changes requested):

query/language/expression.feature
krishnangovindraj: We can do this, but there are tests from ~2 years in reduce.feature that use the `?` on the lhs.
It's also there in the old spec, so I assume that's where it comes from: https://github.com/typedb/typedb/blob/master/docs/archived/spec_old.md#theory-function-semantics 

I think it's great for readability as well so you don't have to go and check whether a return in optional or not. I've not banned it since it was previously allowed, but it's not required anywhere anymore.

query/language/expression.feature
krishnangovindraj: So that I have one row with a value and one without to see if something slips through and breaks.

query/language/fetch.feature
krishnangovindraj: _"require non-empty values"_ isn't the most descriptive either. It's unclear what happens to empty values. "to match only non-empty values"

_"cannot be used without handling the empty value"_ is presumably true only till we have optional arguments where empty values are allowed.

So I've gone for: 
_"The optional variable '{variable}' was used in a context where optionals are not permitted. Use `isset` in the pattern to match only non-empty values. Expressions can use `?` to propagate the empty value to the result."_

query/language/expression.feature
```diff
@@ -2410,3 +2410,101 @@ Feature: TypeQL Query with Expressions
...
       | std::math::max | 10.2dec | 13.5dec | decimal     | 13.5dec |
+
+
+  ###############
+  # OPTIONALITY #
+  ###############
+
+  Scenario: Expressions may return optional values.
+    Given connection open read transaction for database: typedb
+    When typeql read query; fails with a message containing: "The optional variable 'x' was used in a context where optionals are not permitted"
```
flyingsilverfin: if we can advise using the `?` optionality propagation because it's in an expression that would be great

query/language/expression.feature
```diff
@@ -2410,3 +2410,101 @@ Feature: TypeQL Query with Expressions
...
+    """
+    Then uniquely identify answer concepts
+      | x               | y                 |
+      | value:integer:5 | value:integer:13  |
+      | none            | none              |
+
+
+  Scenario: Short-circuits cannot be used where optional results are not allowed
+    Given connection open read transaction for database: typedb
+    Then typeql read query; fails with a message containing: "The optional variable '_anonymous' was used in a context where optionals are not permitted"
```
flyingsilverfin: why isn't this saying `$x` in the error?

query/language/fetch.feature
```diff
@@ -880,14 +880,25 @@ Feature: TypeQL Fetch Query
 
 
   Scenario: fetch can handle optional objects
+    Then typeql read query; fails with a message containing: "The optional variable 'e' cannot be used in a fetch entry without '?'. Use '?' to propagate the empty optional into the fetch entry."
```
flyingsilverfin: there's one! great

flyingsilverfin (approved):
Approved w a few questions to consider

PR #449, opened 2026-09-16
Introduce if statements in write stages

krishnangovindraj:
## Usage and product changes
Introduces if statements in write stages


query/language/insert.feature
```diff
@@ -2880,6 +2900,28 @@ Parker";
     """
 
 
+  Scenario: nested if blocks in insert are disallowed
```
flyingsilverfin: For now eh!

query/language/insert.feature
```diff
@@ -2897,3 +2939,59 @@ Parker";
...
+      };
+    """
+    Then transaction commits
+
+    Given connection open write transaction for database: typedb
+    Then typeql write query; fails with a message containing: "The language feature is not yet implemented"
+    """
+    match $p isa person, has age $age;
+    insert
+      if { $age < 35; } then {
```
flyingsilverfin, line 2968: This is hella cool. we should definitely update docs and write a blog about it

query/language/put.feature
```diff
@@ -733,6 +733,7 @@ Feature: TypeQL Put Query
 
 
   Scenario: nested try blocks in put are disallowed
+    # Nested if blocks are allowed
```
flyingsilverfin, line 736: note this contradicts a test previously (where i left a note)

query/language/variables.feature
```diff
@@ -342,6 +341,12 @@ Feature: TypeQL Variable binding tests
       let $x = none_integer(); # Add ? when enforced
       let $y = $x;
     """
+    When typeql read query; fails with a message containing: "The optional variable 'x' was used in a context where optionals are not permitted"
+    """
+    match
+      let $x = none_integer(); # Add ? when enforced
```
flyingsilverfin, line 347: I believe i had made a note somewhere about better error messages?

flyingsilverfin: sheesh i can't find em. I think the error messages you wrote throughout can be more self-guiding and better worded :)

krishnangovindraj: I use substrings of the error in BDD so we don't have to update them for every little re-wording but we still ensure it's the right error being thrown. The full message is in core.

query/language/insert.feature
krishnangovindraj: It's ignored because it doesn't work

query/language/insert.feature
krishnangovindraj: No, I've allowed them now.

query/language/put.feature
krishnangovindraj: Good catch, I updated that.

query/language/variables.feature
krishnangovindraj: https://github.com/typedb/typedb-behaviour/pull/448#discussion_r4029707978

query/language/insert.feature
krishnangovindraj: Right I remember why now. It's because I can't tell the difference between a comparison in the body and one in the insert statement.
It's doesn't have to be a massive change, but probably better as one that gives write stages their own path from IR down.
Maybe better as a follow up PR once these are in master.

query/language/variables.feature
flyingsilverfin: ` "The optional variable '{variable}' was used in a context where optionals are not permitted. Use isset in the pattern to match only non-empty values. Expressions can use ? to propagate the empty value to the result."` is a good message!

flyingsilverfin (approved):

flyingsilverfin:
Can you make an issue in core to finish implementing generalized conditional writes pls, if you haven't got it up next on your pipeline

flyingsilverfin (approved):
Existing comments on error messages if we can would be ideal but approval stands

